PR #4, opened 2021-06-01
set up basic package infrastructure

### LINKED ISSUE #1: add basic package structure
see title

ref info: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_sprints/backlog/Stress%20Testing/2DegreesInvesting/2021_11

jacobvjk:
closes #1 

First commit to new public stress test repo to be. No content yet, just the skeleton of the repo and package

This PR:
- creates package r2dii.climate.stress.test
- adds corresponding LICENSE, DESCRIPTION and README files
- adds relevant lines to .gitignore and .Rbuildignore
- adds GH Actions


.github/workflows/pr-check-file-encodings.yaml
```diff
@@ -0,0 +1,21 @@
+name: check file encodings in PR
```
maurolepore, line 1: This workflow seems broadly useful. Would it make sense to add it to [r2dii.usethis/inst/templates](https://github.com/2DegreesInvesting/r2dii.usethis/tree/master/inst/templates) so we can add it wherever is needed with something like this?:

    usethis::use_template("pr-check-file-encondings.yaml", package = "r2dii.usethis")


Beyond the convenient way to add it, there is also the benefit of defining a central place to maintain it.  @cjyetman, what the current home of this workflow?

maurolepore (approved):
LTGM

.github/workflows/pr-check-file-encodings.yaml
cjyetman: https://github.com/2DegreesInvesting/resources/issues/53
🤣

cjyetman: maybe this? https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization#creating-a-workflow-template

cjyetman: I started something related to this here https://github.com/2DegreesInvesting/.github/pull/6

Issue #5, opened 2021-06-02
Re-introduce improved litigation risk documentation


jacobvjk:
clarify open points from litigation risk documentation of the old repo before adding

Issue #3, opened 2021-06-01
flexible data path function points to r2dii.stress.test.data by default


jacobvjk:
The new repo needs to allow the user to read in data from directories outside of the repository

as such the data path function needs to be overhauled.
It should point to the r2dii.stress.test.data repo by default, assuming that repo is in the same directory as r2dii.climate.stress.test
The user should be able to point the data path to any other place if necessary

for reference: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/503

PR #6, opened 2021-06-02
2 copy old repo

### LINKED ISSUE #2: copy content from StressTestingModelDev
Everything except the `data-raw` folder from `StressTestingModelDev` repo should be copied into this repo 1 to 1.

for reference: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/487

jacobvjk:
closes #2 

relates to https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/487

Beside copying the contents of [StressTestingModelDev](https://github.com/2DegreesInvesting/StressTestingModelDev), the following things should be noted:
- the entire `data-raw` directory was not copied so that no data is published that we cannot distribute
- I changed one test to `testthat::expect_equivalent` instead of `testthat::expect_equal()` because there were differences in attributes that do not matter for `test-exclude_compnies.R`
- I added `.Rprofile` to `.gitignore` as there are likely future use cases for that, when pointing to external repos/directories for the data
- The code is expected to pass all tests in this state, but actually running the analyses will fail because the `data_path()` does not point to anything meaningful yet. A function will be adapted to handle this properly in the follow up issue #3 
- I removed the documentation of the litigation risk model from the README file temporarily, because some clarifications need to be made before publishing the document. follow up #5


cjyetman (approved):
thanks for making me aware of this, but.... is there something specific you would like me to review here?.... I think obviously this is not the type of PR one can feasibly review the whole thing

Your description/comments/links make reasonable sense to me and are appreciated.

maurolepore (approved):
LGTM as the review seems to be more a formality than a request for feedback -- as CJ points.

* Notice `?testthat::expect_equivalent()`:



* I'm curious why there is was an .Rprofile at the project-level. While that is valid, more common is to have an .Rprofile at the user level -- in the home directory. What kinds of things did that project .Rprofile contain?

jacobvjk:
@maurolepore it's interesting you point this out about the deprecation of expect_equivalent. I used the suggested alternative before and it failed here. Seems odd..
as for the .Rprofile: I can move that to the home directory as well. I don't think any user will be required to use it, so I'm happy to go along with what's conventionally done on that end

jacobvjk:
So I got this when running test-exclude_companies.R

```
>   testthat::expect_equal(
+     input_not_excluded,
+     output_not_excluded,
+     ignore_attr = TRUE
+   )
Error: `input_not_excluded` not equal to `output_not_excluded`.
Attributes: < Component “class”: Lengths (4, 3) differ (string compare on first 3) >
Attributes: < Component “class”: 3 string mismatches >
```

Following what Hadley described over [here](https://github.com/r-lib/testthat/issues/1241) for a similar problem, I solved it like this:

```
>   testthat::expect_equal(
+     unclass(input_not_excluded),
+     unclass(output_not_excluded),
+     ignore_attr = TRUE
+   )
```

jacobvjk:
I stand corrected.. 
now I still get this error:
```
Error: unclass(input_not_excluded) not equal to unclass(output_not_excluded).
Attributes: < Length mismatch: comparison on first 1 components >
```
which can be solved using:

```
> attr(input_not_excluded, "spec") <- NULL
>   
>   testthat::expect_equal(
+     unclass(input_not_excluded),
+     unclass(output_not_excluded)
+   )
```
which in turn leads to me to question what this `ignore_attr = TRUE` is really good for.. ??

maurolepore:
@jacobvjk this reprex suggests the problem is that you have not setup testthat edition 3  (`usethis::use_testthat(edition = 3)`).

``` r
library(testthat)

x <- 1L
attributes(x) <-list(some_attribute = "test_x")
y <- 1L
attributes(y) <-list(some_attribute = "test_y")
waldo::compare(x, y)
#> `attr(old, 'some_attribute')`: "test_x"
#> `attr(new, 'some_attribute')`: "test_y"


# good
test_that("expect_equal() is sensitive to `ignore_attr`", {
  local_edition(3)  
  testthat::expect_equal(x, y, ignore_attr = TRUE)
})
#> Test passed 🥇

# bad
test_that("expect_equal() is sensitive to `ignore_attr`", {
  local_edition(2)  
  testthat::expect_equal(x, y, ignore_attr = TRUE)
})
#> ── Failure (<text>:12:3): expect_equal() is sensitive to `ignore_attr` ─────────
#> `x` not equal to `y`.
#> Attributes: < Component "some_attribute": 1 string mismatch >


# See https://testthat.r-lib.org/articles/third-edition.html
# See ?usethis::use_testthat(edition = 3)
testthat::edition_get()
#> [1] 3
desc <- readLines("DESCRIPTION")
desc[grepl("Config", desc)]
#> [1] "Config/testthat/edition: 3"
```

<sup>Created on 2021-06-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

maurolepore:
RE .Rprofile, I only pointed that out to check your .Rprofile does not include stuff that should instead live in an analysis script. Here is what I have in [my own .Rproile](https://github.com/maurolepore/dotfiles/blob/master/Rprofile). For details see https://rstats.wtf/.

jacobvjk:
.Rprofile: no, it definitely does not contain stuff for a script.
testthat: and you were right about the testthat version. I am a bit surprised, because I am on testthat 3 in the old repo. Naively I assumed that this would also be the case for other new projects... so, will fix this with the next PR

maurolepore:
The explicit commitment to testthat 3 is indeed weird but "by design", else too many packages would have broken when testthat 3 was released.

PR #7, opened 2021-06-03
3 flexible data path

### LINKED ISSUE #3: flexible data path function points to r2dii.stress.test.data by default

jacobvjk:
closes #3 

This PR:
- adds a function that points to the r2dii.stress.test.data repo, if provided the correct environment variable or option
- updates the README to explain required steps to the user
- adapts the work flows to source the function and use it to set the relevant data paths
- adds comments to update the path for the web scripts, once we have clarity on how to pass them


stress_test_model_loan_book.R
```diff
@@ -72,10 +71,8 @@ price_data_version <- cfg_st$price_data_version
 calculation_level <- cfg_st$calculation_level
 company_exclusion <- cfg_st$company_exclusion
 
-# set internal data location, should potentially be moved as this is directly in repo
-# data_location <- paste0(working_location,'data_updated_tool/')
-data_location <- file.path(working_location, data_path())
-# if this is moved out of repo, possibly use set_analysis_inputs_path()
+data_location <- file.path(get_st_data_path(), data_path())
+
 data_location <- ifelse(
```
cjyetman, line 76: L76-80 seem like they will circumvent this new fanciness if this is run on the server, or more precisely when `twodii_internal == FALSE`.... it's that expected?

stress_test_model_eq_cb.R
```diff
@@ -68,10 +67,8 @@ price_data_version <- cfg_st$price_data_version
 calculation_level <- cfg_st$calculation_level
 company_exclusion <- cfg_st$company_exclusion
 
-# set internal data location, should potentially be moved as this is directly in repo
-# data_location <- paste0(working_location,'data_updated_tool/')
-data_location <- file.path(working_location, data_path())
-# if this is moved out of repo, possibly use set_analysis_inputs_path()
+data_location <- file.path(get_st_data_path(), data_path())
```
cjyetman, line 70: when this pattern is used (many places in this PR).... I guess the assumption has to be that `data_path()` will always return a **relative** path... I don't know `data_path()` enough to know if that's a safe assumption

web_tool_stress_test.R
```diff
@@ -71,6 +73,7 @@ company_exclusion <- if (exists("company_exclusion")) company_exclusion else FAL
 
 #### Analysis Parameters----------------------------------------
 # Get analysis parameters from the projects AnalysisParameters.yml - similar to PACTA_analysis
+# TODO: should this remain "working location"? comes from PACTA, so probably okay?
```
cjyetman, line 76: yeah, I'm not sure... I guess we can cross that bridge when we get to it, but yeah, I had the same thought

README.Rmd
```diff
@@ -33,6 +33,23 @@ You can install development version from [GitHub](https://github.com/) with:
 devtools::install_github("2DegreesInvesting/r2dii.climate.stress.test")
 ```
 
+Tu run most of the analyses, you will need additional auxiliary data, such as
+scenario data. The specific requirements are outlined below. **TODO**
+
+Before running the code on a local machine, set the path to the data files
+either as an Environment Variable or an R option using:
```
cjyetman, line 40: might be good to specify explicitly if this var/value is required to be an **absolute** path, **relative** path, or either.... and likewise whether the return value can be expected to be one or both

cjyetman (approved):
generally looks good to me.... made a few comments worth considering

stress_test_model_eq_cb.R
jacobvjk: `data_path()` returns `data-raw/...` by default. `data-raw` can theoretically be changed for something else.. `...` can be an arbitrary number of additional levels and the ultimate file name to be retrieved. so in that sense it is supposed to be used as something returning a relative path within a repo/package/... 
so this should be good as long as the target directory retrieved from `get_st_data_path()` points to the top level of a repo/package or something structured like that.
But I admit I am wondering how reasonable it is to even make two functions out of those.

stress_test_model_loan_book.R
jacobvjk: In a way it is expected, but probably not necessary. Given the script will force any user to set up either an environment variable determining the data location or an option to do that, there is not really any point in allowing users to override that via the config I guess. Best case it's redundant, worst case it leads to errors or unexpected behaviour/results.. Looks like it's time to remove this option

README.Rmd
jacobvjk: fair point

R/get_st_data_path.R
```diff
@@ -0,0 +1,25 @@
...
+#'   "ST_DATA_PATH" cannot be found, the function tries to get the data path
+#'   from options. In that case, provide the name of the option that points
+#'   to the data directory/repo. Default option to get the path is "st_data_path".
+#'
+#' @family miscellaneous utility functions
+#'
+#' @return Character
+#'
+#' @export
+get_st_data_path <- function(envvar = "ST_DATA_PATH",
```
maurolepore, line 16: Example? Maybe reuse the code you write for tests.

stress_test_model_loan_book.R
jacobvjk: so this data_location_ext stuff is removed now. doing that even allowed me to get rid of some confusing and unnecessary paths that I used to set up for nothing in the web tool scripts. should make it a bit easier to figure out the integration once the time comes.

stress_test_model_eq_cb.R
jacobvjk: #8

web_tool_stress_test.R
jacobvjk: #9

maurolepore (approved):
LGTM but I only reviewed the main function. I see other changes that I haven't seen in detail.

See my suggestion for rewrite and tests.

R/get_st_data_path.R
```diff
@@ -0,0 +1,25 @@
...
+                             option = "st_data_path") {
+  if (Sys.getenv(envvar) != "") {
+    return(fs::path(Sys.getenv(envvar)))
+  } else if (!is.null(options(option)[[option]])) {
+    return(fs::path(options(option)[[option]]))
+  } else {
+    stop("Please add data path as envvar or R option")
+  }
+}
```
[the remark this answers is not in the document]
jacobvjk: I was already happy I managed to write a test using withr, but I'll admit this is a lot nicer and the test much more thorough. Will implement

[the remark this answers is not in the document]
jacobvjk: I am changing the second and third tests a bit though, because it can have check() fail locally if the environment variable is set in the .Rprofile
I am doing the following to circumvent the issue and ensure that no environment variable gets matched:

```
test_that("works with `var` set as an R option", {
  path <- "/as/options"
  withr::local_envvar(list(ST_DATA_PATH = NULL))
  withr::local_options(list(ST_DATA_PATH = path))

  expect_equal(get_st_data_path(), path)
})
#> Test passed 😀

test_that("if `var` is unset errors gracefully", {
  withr::local_envvar(list(ST_DATA_PATH = NULL))
  expect_error(get_st_data_path(), "var.*unset")
})
#> Test passed 🎉
```

@maurolepore

PR #25, opened 2021-07-23
884 basel lgd

jacobvjk:
closes ADO 884 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/884

This PR:
- adds two new parameters to `model_parameters.yml`: `lgd_senior_claims` and `lgd_subordinated_claims`. Both are based on the Basel framework per default.
- These parameters are passed to the `company_expected_loss()` function instead of sectoral placeholder LGDs
- as per default, bonds get the `lgd_subordinated_claims` parameter whereas loans get `lgd_senior_claims`
- the `company_expected_loss()` function is adjusted to not merge sectoral lgds but instead create a new variable based on the input from the passed parameter.
- The test is adjusted to use parameter inputs
- The test file with placeholder sectorial LGDs is removed

### DIFF
--- a/.github/workflows/pr-check-file-encodings.yaml
+++ b/.github/workflows/pr-check-file-encodings.yaml
@@ -1,7 +1,8 @@
 name: check file encodings in PR
 on: [pull_request]
 jobs:
-  build:
+  file-encoding:
+    name: file encooding check
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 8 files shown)


jacobvjk:
The encoding check failing is expected, since I removed a file that is not required anymore and this is not covered by the test thus far.

PR #27, opened 2021-08-13
Ado 987

MirijaHa:


MirijaHa:
Closes Task: ADO 978

This tasks removes uncalled functions from repository.
Note: 

- The probably uncalled function _f_ was not removed. Due due the function name it was difficult to rule out with last certainty that it does not get called somewhere
- function _calc_boe_exposures_ is defined twice. As it would create overhead to decide on one and adjust parts of the code relying on the alterantive function definition this is left in for now and will be refactored upon code restructuring in the future.

jacobvjk (approved):
lgtm

PR #30, opened 2021-08-16
1185 new profit calculation

jacobvjk:
closes ADO 1185 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1185

DO NOT MERGE

**Some notes:**
- in principle, this PR implements the new profit function as discussed in the parent PBI of the linked task.
- However, test runs have shown some remaining issues with the underlying data
  - adding a price channel to the profit calculation that applies a carbon tax on production*emission_factor makes sense in theory, but the EFs may be too coarse as of now.
  - also, simply subtracting this from the cashflows in the first step implies that the cashflow calculations cannot simply remain agnostic to absolute price levels as long as relative changes between scenarios are consistent. Subtracting this absolute value based on the emissions and the carbon tax requires the initial cash flow to be based on reasonable absolute levels of prices

TEST STILL OPEN


MirijaHa:
@jacobvjk  a clean repo is a happy repo, how about closing this PR? you can still keep the branch around

PR #33, opened 2021-08-24
Ado 1325

MirijaHa:
This PR closes ADO 1325
it only allow internal mode for now.

Given the clearcut usecase of the function I added no universe but instead tested the possible cases automatically, namely:

- project with configured masterdata, 2dii mode: files get imported
- project without configured masterdata, 2dii mode: descriptive error
- external mode: descriptive error


0_global_functions_st.R
```diff
@@ -180,3 +180,30 @@ check_valid_cfg <- function(cfg, expected_no_args = 3) {
   invisible(cfg)
 }
 
+# create paths to stress test masterdata files
+create_stressdata_masterdata_file_paths <- function(data_prep_timestamp, twodii_internal) {
+  if (!twodii_internal) {
+    stop("Currently cannot provide data files for external mode.")
```
jacobvjk, line 186: should we make a follow up task to decide on how to provide that data and how to handle this here?

0_global_functions_st.R
```diff
@@ -180,3 +180,30 @@ check_valid_cfg <- function(cfg, expected_no_args = 3) {
   invisible(cfg)
 }
 
+# create paths to stress test masterdata files
+create_stressdata_masterdata_file_paths <- function(data_prep_timestamp, twodii_internal) {
```
jacobvjk, line 184: in general: why not put this into the `R/` directory?

jacobvjk (commented):
just two questions

0_global_functions_st.R
MirijaHa: I would add that to the bigger refactoring instead of making it an own ticket now, since there are also a few other small things that do not work in internal mode yet for the loanbook part

0_global_functions_st.R
MirijaHa: True, how about adding it to utils.R?

jacobvjk: seems like a good place

jacobvjk (approved):
lgtm

PR #42, opened 2021-09-03
remove GECO

MirijaHa:
yesterday we agreed I can remove GECO and CPS scenarios. However it turned out CPS is very intermingled in the code, so for now I only deleted GECO, just to get rid of some lines.


jacobvjk (approved):
lgtm

PR #44, opened 2021-09-09
Ado 1639 prewrangled

MirijaHa:
I applied the bulk of prewrangling from the code to the masterdata sets, saved them under a new name, and removed related steps from code. This went along with a very noticable speed improvements of the scripts and also the size of the data decreased (factor 10 in R, up to factor 300 in dropbox).
I checked that datasets used in the scripts subsequently are identical to those used in master version.
I also harmonised the namings a bit and checked that no more unneeded cols are dragged along.

In this form the code should be easy to review, but I would like to suggest a few more improvements as follow up:

- if possible can we simplify the file locations to not depend on a project specific timestamps. This would also make it easy to then drop function create_stressdata_masterdata_file_paths
- There is NaNs in the data. I do not know how they are dealt with later, but it might make sense to put some added effort into tidying this in the saved files
- later in the scripts some trimming/overwriting of net_profit margins happens to the dataset. Let us please discuss if we assume this will be a permanent solution (--> add it in prewrangling), or if that is subject to change (--> move steps to a function)

### DIFF
--- a/R/read_company_data.R
+++ b/R/read_company_data.R
@@ -1,15 +1,23 @@
 #' Read in company finacial data processed from eikon exports and AR master data
 #' that contain information on multiple credit risk inputs and company
-#' production plans.
+#' production plans, aggregated to the ticker/technology/year level for
+#' `asset_type` bonds and aggregated to the company/technology/year level for
+#' `asset_type` equity.
 #'
 #' @param path A string that points to the location of the file containing the
 #'   company financial data.
+#' @param asset_type A string indicating if company data are for analysis for
+#'   bond or equity.
 #'
 #' @family import functions
 #'
 #' @export
-read_company_data <- function(path = NULL) {
+read_company_data <- function(path = NULL, asset_type) {
   path %||% stop("Must provide 'path'")
+
+  if (!asset_type %in% c("bond", "equity")) {
+    stop("Invalid asset type.")
+  }
 
   valid_input_file_path <- file.exists(file.path(path))
   stopifnot(valid_input_file_path)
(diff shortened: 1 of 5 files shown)


R/read_company_data.R
```diff
@@ -1,49 +1,42 @@
...
 #'
 #' @family import functions
 #'
 #' @export
-read_company_data <- function(path = NULL) {
+read_company_data <- function(path = NULL, asset_type) {
+
   path %||% stop("Must provide 'path'")
 
+  if (!asset_type %in% c("bonds", "equity")) {
```
jacobvjk, line 19: should this at some point include loans as a valid asset type? should we then have another place that defines valid asset types?

R/read_company_data.R
```diff
@@ -1,49 +1,42 @@
...
-    "overall_data_type_leverage_s_avg",
-    "overall_data_type_asset_volatility_s_avg",
-    "overall_data_type_asset_drift_s_avg",
-    "overall_data_type_weighted_average_cost_of_capital_percent_s_avg",
-    "overall_data_type_ebitda_ltm_1_usd_s_avg",
-    "overall_data_type_ebitda_margin_percent_ltm_1_s_avg"
+    "company_name", "company_id", "ald_sector", "technology", "year",
+    "ald_production_unit", "ald_emissions_factor_unit", "ald_emissions_factor",
+    "pd", "profit_margin_preferred", "profit_margin_unpreferred", "leverage_s_avg",
+    "asset_volatility_s_avg", "ald_production"
```
jacobvjk, line 33: good for now, but we should discuss with Moritz how to rename some of these more generically.. 
e.g. `leverage_s_avg` is a name from EIKON. in theory this info could come from anywhere, and simply calling it leverage might be preferable in that case. So in the internal prep we still know this name, but for external use, we simplify things

R/utils.R
```diff
@@ -109,8 +109,8 @@ create_stressdata_masterdata_file_paths <- function(data_prep_timestamp, twodii_
   path_parent <- path_dropbox_2dii("PortCheck", "00_Data", "07_AnalysisInputs", data_prep_timestamp)
 
   paths <- list(
-    "stress_test_masterdata_debt.rds",
-    "stress_test_masterdata_ownership.rds",
+    "prewrangled_financial_data_bonds.rds",
+    "prewrangled_financial_data_equity.rds",
     "stress_test_masterdata_credit.rds"
```
jacobvjk, line 114: why not also update it for credit? if not let's set up a follow up task

jacobvjk (commented):
In principle this seems great, some minor comments though

R/utils.R
MirijaHa: I did not update if for credit yet, as there is no prewrangling code around in the code yet, so I did not know how to prewrangle it.
If it is exactly the same steps as for bonds and eq I am happy to do that and to also extend the valid asset types in read_company_data. If not I would prefer to put both aspects in a follow up :)

R/read_company_data.R
jacobvjk: maybe for this one first, for other data types Moritz' input will not be needed

R/utils.R
jacobvjk: fair point

R/read_company_data.R
jacobvjk: see below

MirijaHa:
last question, is there somewhere I shall save the cleaning files I used? i can also keep them locally till  overall convention is determined.

jacobvjk:
I would place them in the same directory as the previous stress_test_master_data files for now

PR #47, opened 2021-09-15
hot fix: explicitly filter out etp2017 - power sector comboination fr…

jacobvjk:
…om pacta data

closes ADO 1868 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1868

This PR
- makes a hotfix that removes the combination of pacta resultswith scenario source ETP2017 and ald_sector power in the web script 
- keeping this combination will introduce duplicates, as the code cannot distinguish it from the WEO2019 power sector
- it seems like in the past this was not a problem, because that combination of data was not passed


MirijaHa (approved):
amazing

PR #48, opened 2021-09-15
use testing project per default

MirijaHa:
This closes ADO_1818
Project ST_TESTING was set up based on TRISK project, but all data that is not necessary for the (currently only) supported ec_cb workflow was removed to facilitate getting an overview.


jacobvjk (approved):
lgtm

PR #51, opened 2021-09-21
Check inputs

MirijaHa:


MirijaHa:
Relates to ADO 1899.

- added functions to run checks for duplicate (all and selected cols) as well as missing combinations on data
- I did not apply this to any data sets yet but would like to already have them in the package for interactive use and possible add them to selected input data

R/utils.R
```diff
@@ -155,3 +155,86 @@ validate_data_has_expected_cols <- function(data,
...
+}
+
+#' Identify and report missing value combinations
+#'
+#' Checks if all level combinations of `composite_unique_cols` are in`data` and
+#' throws a warning on missing combinations.
+#' NOTE:
+#' 1. a combination of all levels is not neccesarily required/useful, make sure
+#' to use function only in adequate context.
+#' 1. combiantions of too many columns/values may exceed memory size.
```
jacobvjk, line 197: 2. not 1.?

R/utils.R
```diff
@@ -155,3 +155,86 @@ validate_data_has_expected_cols <- function(data,
...
   stopifnot(data_has_expected_columns)
 }
+
+#' Checks data for missings and duplicates
+#'
+#' Applies consistency checks to data concerning the combinations of columns
+#' that should be unique in combination. In concrete:
+#'
+#' 1. it is checked if there are duplicate rows.
+#' 1. it is checked if there are duplicate rows on `composite_unique_cols`.
```
jacobvjk, line 165: 2. not 1.?

R/utils.R
```diff
@@ -155,3 +155,86 @@ validate_data_has_expected_cols <- function(data,
...
+#' .
+#'
+#' @inheritParams report_all_duplicate_kinds
+#'
+#' @return NULL
+#' @export
+report_missing_col_combinations <- function(data, composite_unique_cols) {
+
+  all_combinations <- data %>%
+    tidyr::expand(!!!dplyr::syms(composite_unique_cols))
```
jacobvjk, line 207: just a question: would you prefer to use `!!!dplyr::syms()` in the future? I Have thus far namespaced it with `rlang`, where I believe the function originates from, but I guess they will keep it in sync in `dplyr` as well

jacobvjk (commented):
some tiny documentation comments, otherwise makes sense to me

R/utils.R
MirijaHa: esoteric roxygen stuff. The second 1 will be translated to 2. in the manual. This way it is easier to change order or add elements

R/utils.R
MirijaHa: Since we would rather get rid of rlang  as explicit dependency I did not want explicitly say rlang::syms. At this, rlang syms will be equal or ahead of dplyr syms, so getting it from dplyr should be a bit safer (dplyr must have decided to upgrade..)

jacobvjk: seems reasonable, will keep in mind for the future

jacobvjk (approved):

PR #53, opened 2021-09-23
Ado 1920 split

MirijaHa:


MirijaHa:
PR separates EQ from CB workflows. I ran checks that scripts still function and results are identical to results of master version (set seed for bonds workflow).

jacobvjk (approved):
lgtm
maybe we can even remove the distinction between EQ and CB at some point and just allow the user to say if they want to calculate credit risk a well or not. I think in terms of data structure the two cases should otherwise be identical

PR #55, opened 2021-09-23
Ado 1518 change checks

MirijaHa:
Added 1 script per eq, cb and cl to the QA scripts which helps to easily compare all results generated with 2 different version.
Scripts are currently unnecessarily repetitive, I will simplify them once there is a proper main function.
I tested them for eq and cb, I could not test the recalcs for cl since script currently does not run but compared result comparison there.


jacobvjk (approved):
lgtm

PR #56, opened 2021-09-23
remove unneeded file

MirijaHa:
removes imports of carbon tax data.
I left the importing function in place though since we will need it not too long from now likely


jacobvjk (approved):
lgtm

PR #60, opened 2021-09-29
879 use eikon data loanbook workflow

jacobvjk:
closes ADO 879: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/879

This PR:
- uses the prewrangled stress test master data for loans (`financial_data_loans`)
- it adds further preparation steps for the `net_profit_margin` (corresponding to CB workflow)
- it joins the `financial_data_loans` with the loan book in the main calculation to add the new profit margins and the variables for the Merton crediit risk model (corresponding to CB workflow)
- it temporarily merges this `financial_data_loans` to the portfolio by joining on company names only (the company names are transformed to lower for that)
- this should be replaced for `company_id` joining later on, once P4B provides `company_id` as an output variable (see: https://github.com/2DegreesInvesting/r2dii.match/issues/375)
- removes loading the former net profit margins into the environment
- removes the call of `join_net_profit_margins()` from the main calculation (comes directly through the `financial_data_loans`)
- adapts `write_results()` so that it will accept loan book stress test results
- use `write_results()` in loan workflow
- fix typo in expected loss output preparation


R/write_results.R
```diff
@@ -28,7 +28,7 @@ write_results <- function(
   level %||% stop("Must provide 'level'")
   investorname %||% stop("Must provide 'investorname'")
```
MirijaHa, line 29: export technically works, just a quick content question. I noticed that on variable VaR_tech_company there is NaNs, positive and negative numbers, is that expected?

stress_test_model_loan_book.R
```diff
@@ -530,9 +523,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
       )
   }
 
+  loanbook_annual_profits <- loanbook_annual_profits %>%
+    # ADO 879: removed company id from join, but should be re-introduced later on
+    left_join(financial_data_loans, by = c("company_name", "ald_sector", "technology", "year"))
```
MirijaHa, line 528: what is the reasoning behind using a left join here, and what is the hanlding for the resulting NAs on e.j. column pd?

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
       year == start_year,
       technology %in% technologies,
       scenario_geography == scenario_geography_filter
+    ) #%>%
+    # distinct(investor_name, portfolio_name, company_name, ald_sector, technology,
```
MirijaHa, line 564: please remove commented out code

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+    #          scenario_geography, year, plan_carsten, plan_sec_carsten, term,
+    #          PD_0)
+
+  financial_data_loans_pd <- financial_data_loans %>%
+    select(company_name, company_id, ald_sector, technology, pd) %>%
+    distinct(across(everything()))
```
MirijaHa, line 570: distinct(across(everthing)) is quite problematic. 1) it overshadows data consistency problems and 2) we might still keep duplicates on composite unique columns. How about instead filtering by star_year and then calling report_all_duplicates instead?

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+
+  plan_carsten_loanbook <- plan_carsten_loanbook %>%
+    # ADO 879: removed company id from join, but should be re-introduced later on
+    left_join(financial_data_loans_pd, by = c("company_name", "ald_sector", "technology"))
+  # TODO: what to do with entries that have NAs for pd?
```
MirijaHa, line 575: is this still open?

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+    #          scenario_geography, year, plan_carsten, plan_sec_carsten, term,
+    #          PD_0)
+
+  financial_data_loans_pd <- financial_data_loans %>%
+    select(company_name, company_id, ald_sector, technology, pd) %>%
+    distinct(across(everything()))
+
+  plan_carsten_loanbook <- plan_carsten_loanbook %>%
+    # ADO 879: removed company id from join, but should be re-introduced later on
+    left_join(financial_data_loans_pd, by = c("company_name", "ald_sector", "technology"))
```
MirijaHa, line 574: again why left join, how are resulting NAs dealt with? Also I noticed that the join increases the numbers of rows, this is cause by duplicates on RHS data on c("company_name", "ald_sector", "technology") and likely not correct

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+  # TODO: what to do with entries that have NAs for pd?
+  # TODO: kick out NAs and record the diff
+  loanbook_annual_profits <- loanbook_annual_profits %>%
```
MirijaHa, line 577: actually, since i noticed this for for equity_pacta_result also,  when is company id na while name is there? Also, would that not created problems in earlier steps, e.g. in any grouping operation that uses company_id?

stress_test_model_loan_book.R
```diff
@@ -547,10 +560,29 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+
+  plan_carsten_loanbook <- plan_carsten_loanbook %>%
+    select(
+      investor_name, portfolio_name, company_name, ald_sector, technology,
+      scenario_geography, year, plan_carsten, plan_sec_carsten, term, pd
     ) %>%
-    distinct(investor_name, portfolio_name, company_name, ald_sector, technology,
-             scenario_geography, year, plan_carsten, plan_sec_carsten, term,
-             PD_0)
+    distinct_all()
```
MirijaHa, line 585: again, i would rather suggest to add a filtering and a checking for duplicates step.

MirijaHa (changes requested):
Handlings is consistent to bonds workflow. However I noticed some robustness issues, I would suggest to have a closer look at them, and maybe also adjusting them for cb and eq to keep consistency.

R/write_results.R
jacobvjk: positive and negative is definitely expected. NaN is something I will check and should probably be handled differently even if the calculation were using correct inputs.

stress_test_model_loan_book.R
jacobvjk: This is a good example where there technically is not need to keep the NA rows. But if we drop them, we may want to think about adding them to some json output column/file that records what is kicked out where and why...

stress_test_model_loan_book.R
jacobvjk: there are a bunch of TODOs commented into various parts of the loans workflow. This was essentially copied here from elsewhere. I would rather address them in dedicated tasks, because I feel they don't really have anything to do with the workflow fundamentally passing all variables to their respective use cases in the same way that the bonds work flow does that. But they should be addressed before the release

stress_test_model_loan_book.R
jacobvjk: This says NAs for PD not ID.. (which essentially comes back to the left join question)

on the id question: there should not be any companies with a name but no id. I think I may have seen such a case in some raw data file (probably financial data) at some point, but they probably do not get carried through to this stage of the code. I realise this is all pretty vague, but in the end this would best be handled at the import

stress_test_model_loan_book.R
MirijaHa: how about as a short term solution dropping them (because keeping them just adds complication and makes understanding things harder) and adding a log line (simply use cat) how many rows where dropped. For proper handling i.e. returning affected cases in some form a follow up ticket can (should) be created)

stress_test_model_loan_book.R
MirijaHa: ok, resolving issue

stress_test_model_loan_book.R
MirijaHa: ok, resolving because not relevant here.

stress_test_model_loan_book.R
jacobvjk: sounds good

jacobvjk:
@MirijaHa there is actually an issue with using the inner_join of the portfolio and the financial data, being that it kicks out the years between the production forecast from the ald (5 years) and the extended scenario trajectories (up until 2040).. we will need to rethink what a good solution to this looks like

jacobvjk:
For the time being, the problem can be solved by filtering the financial data to the start year of the analysis. The only variables that would change over time are the emissions factors and production data, which we will only need from this dataset, once we decouple from pacta

R/write_results.R
MirijaHa: any updates on this? :)

jacobvjk: I found a number pof things that were contributing to issues, but some are still open. Firstly, the inner join on the same columns was problmeatic, because "year" kicked out required rows (my initial reason for using the left_join I guess). I solved this by filtering on the start year before the join, as all the financials are constant over time anyways. I also removed the production and EF columns from the masterdata before joining as they are not required here. I think that part should be fine now.

stress_test_model_loan_book.R
jacobvjk: Removed the left_join here as well.

stress_test_model_loan_book.R
jacobvjk: plan_carsten turned out to produce duplicates because I did not filter on any scenario and the variable (as it is a planned variable) is of course identical across scenarios. So I added a filter for the target scenario further up an removed this

stress_test_model_loan_book.R
jacobvjk: resolved by filtering on start year before the loop

R/write_results.R
MirijaHa: I checked again, and it turns out there are still 44 rows with NaNs for VaR_tech_company and tech_company_value_change (always technology oil) as well as 539 rows with NA on production_shock_perc. Again, I guess you can assess better if that is a problem :)

jacobvjk: The latter is related to pivoting the empty shocks from the shock scenario longer for all companies. It's always NA. This will change with https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1618
The former are the companies with net_profit_margin = 0.
This will be handled appropriately with the follow up tasks [ADO 1677](https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1677) and [ADO 2180](https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2180)

MirijaHa:
> @MirijaHa there is actually an issue with using the inner_join of the portfolio and the financial data, being that it kicks out the years between the production forecast from the ald (5 years) and the extended scenario trajectories (up until 2040).. we will need to rethink what a good solution to this looks like

ist this still up to date?

jacobvjk:
@MirijaHa no, it is solved by filtering on `start_year` only. The financial data are stable over time and the only variation that we have in that financial data set thus far was due to it including production and emission factor data for each company. Since we are only interested in using the financial part of the data here, I removed the production related ones. This way, if `"year"` is not joined on, the financial data will populate all remaining rows in the loan book / portfolio after the `inner_join`

stress_test_model_cb.R
```diff
@@ -324,6 +324,14 @@ financial_data_bonds <- financial_data_bonds %>%
   dplyr::rename(
     debt_equity_ratio = leverage_s_avg,
     volatility = asset_volatility_s_avg
+  ) %>%
+  # ADO 879 - remove year and production/EFs to simplify joins that do not need yearly variation yet
+  dplyr::filter(.data$year == .env$start_year) %>%
```
MirijaHa, line 329: I noticed this was not done for EQ workflow, is that on purpose?

stress_test_model_cb.R
```diff
@@ -470,16 +487,16 @@ for (i in seq(1, nrow(transition_scenarios))) {
     # NOTE: this assumes emissions factors stay constant after forecast and prod not continued
     tidyr::fill(
       company_id, pd, net_profit_margin, debt_equity_ratio, volatility,
-      ald_emissions_factor, ald_emissions_factor_unit, ald_production_unit,
+      #ald_emissions_factor, ald_emissions_factor_unit, ald_production_unit,
```
MirijaHa, line 490: please no commenting out. Again I noticed that handling for equity_annual_profits was not adjusted and wonder if that is intended

stress_test_model_cb.R
```diff
@@ -489,28 +506,43 @@ for (i in seq(1, nrow(transition_scenarios))) {
 
   plan_carsten_bonds <- pacta_bonds_results %>%
     filter(
-      year == start_year,
-      technology %in% technologies,
-      scenario_geography == scenario_geography_filter
+      .data$year == .env$start_year,
+      .data$technology %in% .env$technologies,
+      .data$scenario_geography == .env$scenario_geography_filter,
+      .data$scenario %in% .env$scenario_to_follow_ls
```
MirijaHa, line 512: also not done for EQ

stress_test_model_cb.R
```diff
@@ -455,8 +463,17 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
   }
 
+  rows_bonds <- nrow(bonds_annual_profits)
   bonds_annual_profits <- bonds_annual_profits %>%
-    left_join(financial_data_bonds, by = c("company_name", "id" = "corporate_bond_ticker", "ald_sector", "technology", "year"))
+    dplyr::inner_join(
+      financial_data_bonds,
+      by = c("company_name", "id" = "corporate_bond_ticker", "ald_sector", "technology")
+    )
+  cat("number of rows dropped from loan book by joining financial data on
```
MirijaHa, line 472: NP: note that  n rows could also increase after an inner join. However this will soon not be possible anymore once duplicates checks and filterings are applied to all datasets.

stress_test_model_cb.R
```diff
@@ -489,28 +506,43 @@ for (i in seq(1, nrow(transition_scenarios))) {
...
+      by = c("company_name", "id" = "corporate_bond_ticker", "ald_sector", "technology")
+    )
+  cat("number of rows dropped from technology_exposure by joining financial data
+      on company_name, corporate_bond_ticker, ald_sector and technology = ",
+      rows_plan_carsten - nrow(plan_carsten_bonds), "\n")
+  # TODO: ADO 879 - note which companies are removed here, due to mismatch of
+  # sector/tech in the financial data and the portfolio
   # TODO: what to do with entries that have NAs for pd?
-  # TODO: kick out NAs and record the diff
-  bonds_annual_profits <- bonds_annual_profits %>%
```
MirijaHa, line 505: for my understanding, why could this line be dropped?

stress_test_model_eq.R
```diff
@@ -442,11 +442,14 @@ for (i in seq(1, nrow(transition_scenarios))) {
       )
   }
 
+  rows_equity <- nrow(equity_annual_profits)
```
MirijaHa, line 445: We lose a very higher number (> 10000) rows here, Let us hope this is due to either inconsistencies to cb workflow or due to the fact that import of pacta results is currently not really functional.

stress_test_model_eq.R
```diff
@@ -488,8 +490,12 @@ for (i in seq(1, nrow(transition_scenarios))) {
     select(
       investor_name, portfolio_name, company_name, ald_sector, technology,
       scenario_geography, year, plan_carsten, plan_sec_carsten
-    ) %>%
-    distinct(across(everything()))
+    )
+
+  report_duplicates(
```
MirijaHa, line 495: here quite a lot of duplicates are reported, again hopefully due to inconsistencies of faulty import.

MirijaHa (commented):
Mostly small inconsistencies and issues that are hopefully related to the currently faulty import of pacta results. I would suggest we quickly synch on next steps before u address my comments

stress_test_model_cb.R
jacobvjk: we `inner_join` `id` on the `corporate_bond_ticker`, which does not have any `NAs` for bonds.

stress_test_model_eq.R
jacobvjk: I will make the eq workflow consistent with the cb workflow as part of this https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2182 (including adding the Credit Risk calculations). We will know more then, meaining I will go into why this may be happening

MirijaHa (approved):
Splendid

PR #61, opened 2021-09-30
Switch warn stop

MirijaHa:
I added the option that consistency checks throw error on request. I did this because while for interactive use warnings are fine for use in code it should be an error and I am planning to introduce some consistency checks in the next PR.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -238,7 +238,7 @@ report_missing_col_combinations <- function(data, composite_unique_cols, throw_e
 #' @param cols Cols to check for duplicate combinations on.
 #'
 #' @return NULL
-report_duplicates <- function(data, cols) {
+report_duplicates <- function(data, cols, throw_error) {
   duplicates <- data %>%
     dplyr::group_by(!!!dplyr::syms(cols)) %>%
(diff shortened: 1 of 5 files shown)


R/utils.R
```diff
@@ -170,19 +170,28 @@ validate_data_has_expected_cols <- function(data,
...
-  report_duplicates(data = data, cols = names(data))
+  report_duplicates(
+    data = data,
+    cols = names(data),
+    throw_error = throw_error
+  )
 
-  report_duplicates(data = dplyr::distinct(data), cols = composite_unique_cols) # removed duplicates on all columns
+  report_duplicates(
+    data = dplyr::distinct(data),
```
jacobvjk, line 191: just out of curiosity, if applying `distinct()` to an entire dataset, wouldn't `distinct_all()` be the more natural option?

MirijaHa: it would, but that would mean I would have to make function report_duplicates more complicated or duplicate it. This way I achieve the same with using just 1 function for checking and reporting

jacobvjk (approved):
LGTM

PR #64, opened 2021-09-30
Ado 1939

MirijaHa:
This PR fixes bug introduced by not filtering for valid_input on sector exposures. While at it I did a small refactoring and moved calculation of aum to a function.


R/wrangle_and_check.R
```diff
@@ -0,0 +1,31 @@
...
+#'
+#' @param sector_exposures A tibble holding sector_exposures as calculated by
+#'   PACTA.
+#' @param asset_type Type of analyzed asset. Can be bonds or equity.
+#'
+#' @return `sector_exposures` holding only valid rows for `asset_type`
+#' @export
+wrangle_and_check_sector_exposures_eq_cb <- function(sector_exposures, asset_type) {
+
+  if (!asset_type %in% c("Bonds", "Equity")) {
```
jacobvjk, line 15: are you not doing this for loans, because there is no such flag there? (I think it's correct, just checking if this is assumed?)

jacobvjk (commented):
minor comments

R/wrangle_and_check.R
MirijaHa: Excactly, 

- there both is no flag on invalid_input 
- only 1 asset type is included so no need for filtering.

Thus one has to trust that dealing with valid_input (if necessary) was done correctly whereever it is done. However since I will add a check for duplicates we should notice if valid_input filtering was not done.

R/calculate_aum.R
```diff
@@ -0,0 +1,19 @@
...
+#' @return A tibble holding columns investor_name, portfolio_name and
+#'   asset_portfolio_value.
+#' @export
+calculate_aum <- function(sector_exposures) {
+  aum <- sector_exposures %>%
+    group_by(investor_name, portfolio_name) %>%
+    summarise(
+      asset_portfolio_value = sum(valid_value_usd),
+      .groups = "drop_last"
+    )
```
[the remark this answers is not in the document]
MirijaHa: changed to drop

jacobvjk (approved):
lgtm

PR #68, opened 2021-10-06
Ado 1876 next

MirijaHa:


MirijaHa:
- Integrated all Pacta wrangling for EQ and CB in 1 function  
- moved shared hardcodings for user workflows to lookup file
I tested the changes by running the change_check files for all 3 asset types. 
For eq and cl there are no differences in results in comparison to master version, which is expected since no functional change is made. 
For cb I somehow did not manage to repro results simply by setting a seed. Since that was, for unknown reasons, also again the case when comparing runs of the same version I enden up setting term to 1 for testing purposes. This made results reproducable between master and new version.

R/wrangle_and_check.R
```diff
@@ -29,3 +29,58 @@ wrangle_and_check_sector_exposures_eq_cb <- function(sector_exposures, asset_typ
...
+#' @return Wrangled `pacta_results.`
+#' @export
+wrangle_and_check_pacta_results_eq_cb <- function(pacta_results, start_year, time_horizon,
+                                                  scenario_geography_filter, scenarios_filter,
+                                                  equity_market_filter) {
+
+  wrangled_pacta_results <- pacta_results %>%
+    dplyr::filter(!is.na(.data$scenario)) %>%
+    check_scenario_settings(scenario_selections = allowed_scenarios_eq_cb) %>%
+    dplyr::filter(.data$scenario %in% allowed_scenarios_eq_cb) %>%
```
jacobvjk, line 60: this seems like an implicit argument? Or am I missing something here?

R/wrangle_and_check.R
```diff
@@ -29,3 +29,58 @@ wrangle_and_check_sector_exposures_eq_cb <- function(sector_exposures, asset_typ
...
+    # TODO: temporary fix, remove once all scenario data is used from scenario file
+    filter(!(scenario == "ETP2017_NPS" & ald_sector == "Power")) %>%
```
jacobvjk, line 62: wouldn't it be a good time to also namespace and data mask this section?

jacobvjk (commented):
see comments
I don't want to place the extra burden on you, but maybe we can also start to namespace the functions in the main scripts step by step soon?

R/wrangle_and_check.R
MirijaHa: u mean allowed_scenarios_eq_cb? This is available since it is configured in lookup.R I moved a couple of variables (sectors, technologies..) there to not overload the function interface, given the names are very specific I thinkt this should be no problem

R/wrangle_and_check.R
MirijaHa: good point, doing

MirijaHa: done

PR #69, opened 2021-10-06
adjust format to eq and cb

MirijaHa:
Make scenario handling identical to PACTA results for eq and cb via webtool.


jacobvjk (approved):
lgtm

PR #72, opened 2021-10-08
Ado 2261 negatives

ghost:
added check to remove negative trajectories and did a bit of restyling. I would recommend hiding whitespace changes when reviewing

### DIFF
--- a/R/set_tech_trajectories.R
+++ b/R/set_tech_trajectories.R
@@ -483,7 +483,8 @@ filter_negative_late_and_sudden <- function(data_with_late_and_sudden) {
       dplyr::anti_join(negative_late_and_sudden, by = c("company_name", "technology"))
 
     warning(paste0("Removed ", n_rows_before_removal - nrow(data_with_late_and_sudden),
-                   " rows due to negative late and sudden targets."))
+                   " rows because negative production compensation targets were set in late and sudden.
+                   Negative absolute production is impossible"))
 
     if (nrow(data_with_late_and_sudden) == 0) {
       stop("No rows remain after removing negative late and sudden trajectories.")


R/set_tech_trajectories.R
```diff
@@ -464,3 +458,38 @@ calc_late_sudden_traj <- function(start_year = 2020,
...
+#' removed.
+#'
+#' @param data_with_late_and_sudden A tibble containing scenario data with
+#'   projected late and sudden trajectory.
+#'
+#' @return Input tibble with potentially removed rows.
+filter_negative_late_and_sudden <- function(data_with_late_and_sudden) {
+  negative_late_and_sudden <- data_with_late_and_sudden %>%
+    dplyr::filter(.data$late_sudden < 0) %>%
+    dplyr::select(.data$company_name, .data$technology) %>%
```
jacobvjk, line 475: why not save the select line and directly go `dplyr::distinct(.data$company_name, .data$technology)`? same result, less lines...

R/set_tech_trajectories.R
```diff
@@ -464,3 +458,38 @@ calc_late_sudden_traj <- function(start_year = 2020,
...
+
+  if (nrow(negative_late_and_sudden) > 0) {
+    n_rows_before_removal <- nrow(data_with_late_and_sudden)
+
+    data_with_late_and_sudden <-
+      data_with_late_and_sudden %>%
+      dplyr::anti_join(negative_late_and_sudden, by = c("company_name", "technology"))
+
+    warning(paste0("Removed ", n_rows_before_removal - nrow(data_with_late_and_sudden),
+                   " rows due to negative late and sudden targets."))
```
jacobvjk, line 486: maybe a bit more precies: `"... due to negative production targets in late and sudden scenario"` so it's clear we are talking about absolute values

jacobvjk: `"Removed ", n_rows_before_removal - nrow(data_with_late_and_sudden), " rows because negative production compensation targets were set in late and sudden. Negative absolute production is impossible."`

PR #81, opened 2021-10-12
Call simple merton

ghost:
This closes ADO 1671.
Speed is improved by several minutes per run.
I reproed that results for CL do not change.
There is no reason why it should differ but to go absolutely sure I will also run pre post tests for other asset types.


jacobvjk (approved):
LGTM!

PR #87, opened 2021-10-15
Wrangle fin

MirijaHa:


R/wrangle_and_check.R
```diff
@@ -81,3 +81,44 @@ wrangle_and_check_pacta_results <- function(pacta_results, start_year, time_hori
...
+#' Applies custom improvements of `net_profit_margin` and does row and cols
+#' selections as well as filtering.
+#'
+#' @param financial_data A data set of `financal_data`.
+#' @param start_year String holding start year of analysis.
+#'
+#' @return A prewrangled `financal_data` set.
+wrangle_financial_data <- function(financial_data, start_year) {
+  financial_data <- financial_data %>%
+    dplyr::mutate(net_profit_margin = profit_margin_preferred) %>%
```
jacobvjk, line 96: can I suggest to add data masking where it applies?

MirijaHa: You can, and I thought you would, but since in this case it would be quite a lot of work (all those if else conditions need several masking) and there is a good chance we get rid of this alltogether when we move towards CRAN ready version I would rather not :)

PR #92, opened 2021-10-18
2372 stricter checks

ghost:


tests/testthat/test-filter.R
```diff
@@ -0,0 +1,39 @@
...
+  )
+  expect_error(keep_merton_compatible_rows(input_data, stage = "some"), "Invalid")
+})
+
+
+test_that("input remains unaltered if rows are merton compatible", {
+  input_data <- tibble::tibble(
+    debt = c(1, 2),
+    equity_0_baseline = c(1, 2),
+    equity_0_late_sudden = c(1, -1.99),
```
jacobvjk, line 19: I would tend to argue that equity cannot be negative in our calculations. I know this is not what is being checked and filtered here, but the example should probably still use a plausible number. we derive it from the discounted net profits, which are the product of production, price and the net profit margin, all of which are non negative. discounting will also not turn this into a negative value.
Up to you if we want to change that here or we just keep it in mind for further checks

PR #93, opened 2021-10-18
Review left joins

jacobvjk:
closes ADO 1945: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1945

This PR:
- replaces `left_join`s with `inner_join`s where this makes no difference (or is preferable)
- adds comments wo all remaining `left_join`s as to why they were used
- does not yet apply this to the litigation risk workflow, as this would require more refactoring and is not of relevance to the transition risk model
- the check for equality of results passed


MirijaHa (approved):
LGTM and thanks a lot

PR #94, opened 2021-10-19
Switch analysis input

ghost:
Scenario_data is now used directly from data repo.
Change check for bonds passed.
I sneakily added a small fix in the bonds_change_check also.

### DIFF
--- a/R/read_scenario_data.R
+++ b/R/read_scenario_data.R
@@ -6,12 +6,25 @@
 #'
 #' @return A tibble holding scenario data.
 read_scenario_data <- function(path) {
-  scenario_data <- readr::read_csv(path, col_types = "ccccccncn")
+  scenario_data <- readr::read_csv(path, col_types = readr::cols(
+    scenario_source = "c",
+    scenario_geography = "c",
+    scenario = "c",
+    ald_sector = "c",
+    units = "c",
+    technology = "c",
+    year = "d",
+    direction = "c",
+    fair_share_perc = "d"
+  ))
+
   validate_data_has_expected_cols(
     data = scenario_data,
-    expected_columns = c("scenario_source", "scenario_geography", "scenario",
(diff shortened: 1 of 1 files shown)


R/read_scenario_data.R
```diff
@@ -0,0 +1,17 @@
+#' Read in scenario data
+#'
+#' Function reads in scenario data and checks for existence of required columns.
+#'
+#' @param path Path to dir holding scenario data.
+#'
+#' @return A tibble holding scenario data.
+read_scenario_data <- function(path) {
+  scenario_data <- readr::read_csv(path, col_types = "ccccccncn")
```
jacobvjk, line 9: let's make the col_types explicit while we are at it

R/wrangle_and_check.R
```diff
@@ -126,3 +125,24 @@ wrangle_financial_data <- function(financial_data, start_year) {
...
+#'
+#' @param scenario_data A tibble holding scenario data.
+#' @param start_year Start year of analysis.
+#' @param end_year End year of analysis.
+#'
+#' @return A tibble holding wrangled scenario_data.
+wrangle_scenario_data <- function(scenario_data, start_year, end_year) {
+  scenario_data_wrangled <- scenario_data %>%
+    dplyr::rename(source = .data$scenario_source) %>%
+    dplyr::filter(.data$source %in% c("ETP2017", "WEO2019")) %>%
```
jacobvjk, line 141: we need to remember to parameterise these scenarios at some point

jacobvjk (commented):
see comments

R/wrangle_and_check.R
MirijaHa: we sure do

R/read_scenario_data.R
MirijaHa: what do you mean by that?

PR #100, opened 2021-10-20
1943 term input

jacobvjk:
closes ADO 1943: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1943

This PR:
- adds another input argument `term` to the main functions
- this argument defaults to a 2 year maturity and is validated against a range from 1 to 10 (full years)
- the expected loss calculation is adapted so that it only retains the rows that match the input term (changed left_join to inner_join
- the calculations of annual and overall PDs get comments to explain why we use different hard-coded terms in those

**NOTE:**
- results will change compared to the last version, as the expected loss result retains only a fifth of the former rows (the other ones being irrelevant top that output)
- additionally the term is set to 2 as a default, which inevitably changes most PDs compared to the previous version that had a random placeholder


MirijaHa (commented):
just 1 minor comment

PR #126, opened 2021-10-29
fix rm key in apply_filters

jacobvjk:
closes ADO 2562: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2562

This PR:
- fixes a bug in the apply_filter function, that lead to removing a lot more companies from loan books than desirable
- specifically, it adds the company name as an additional identifier to the remove key in apply filters
- previously the company was only identified via the id, which is insufficient for loans, as we do not have company level ids in loan data sets
- the change checks remain identical for eq and cb and all outputs change for cl. This is exactly the intended effect.


MirijaHa (commented):
Am I right in thinking that running change_checks showed

- no changes for eq und cb
- expected changes for cl?

jacobvjk:
> Am I right in thinking that running change_checks showed
> 
> * no changes for eq und cb
> * expected changes for cl?

very much so :)

MirijaHa (commented):
Before I approve:

- have u thought sharply if any other parts of the code could depend on company_name alone and thus are buggy for loans?
- could this somehow be related to the much smaller number of considered sectors in sector exposures that we observed?
- FYI: the issue with NA rows on VaR company as observed for bonds is now also happening for loans.

jacobvjk:
> Before I approve:
> 
> * have u thought sharply if any other parts of the code could depend on company_name alone and thus are buggy for loans?
> * could this somehow be related to the much smaller number of considered sectors in sector exposures that we observed?
> * FYI: the issue with NA rows on VaR company as observed for bonds is now also happening for loans.

- from what I know and can see, there is no other place in the code, where a calculation or transformation is performed solely on the id, without considering the company_name.
- sector_exposures are not related
- the NA rows issue seems to be related to plan_carsten values of 0 in the pacta result files we use as input for company/technology combinations that have positive production values. This is unexpected behaviour to me and will be reviewed as part of the general methodology review task https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2560

MirijaHa (approved):
marvelous

PR #127, opened 2021-10-29
default credit type outstanding and use credit type as input to p4b calculation

jacobvjk:
closes ADO 2561: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2561

This PR:
- sets the default `credit_type` for preparing the p4b data in the loans workflow to `"outstanding"`, as this is the more commonly provided format and way to think about credit risk
- additionally I found a bug in the loans prep function. The function used to not pass the credit type information to the actual r2dii.analysis::target_market_share calculation, so that in the past it would always have calculated on the credit limit. This is now fixed
- The sample results for the loans workflow will change the next time the prep function is run. Nothing else is affected


MirijaHa (commented):
will approve when merged with master, please make sure to update the data as prepped with this workflow in the project folders also.

MirijaHa (approved):

jacobvjk:
data is updated

PR #132, opened 2021-11-01
EXPERIMENTAL test if CI works with sessionInfo v1.1.1

maurolepore:
This is an experiment to try confirm the issue why CI is failing is the recent (Oct 31st) release of sessionInfo.


maurolepore:
My experiment doesn't work because the failure is in a premade action that errors before this workflow reaches the point where I install an older version of sessionInfo.

I abandon the experiment, hoping the r-lib team will fix things soon.

PR #134, opened 2021-11-01
Address R CMD warnings and notes

maurolepore:
This PR addresses a couple of warnings/notes I see on 
R CMD check. This leaves only one note of this type:

```
  set_project_parameters: no visible binding for '<<-' assignment to
    ‘inc_stresstest’
```

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -24,3 +24,4 @@
 ^wip_data_qa\.R$
 ^\.github$
 ^\.Renviron$
+^vignettes/articles$
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ License: file LICENSE
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 Imports: 
     config,
     dplyr,
--- a/R/calc_survival_probability_merton.R
+++ b/R/calc_survival_probability_merton.R
@@ -2,8 +2,8 @@
 #'
 #' Function calculates survival probability for a maturity based on a structural
 #' Merton model.
-#' For details on implementation please compare [CreditRisk::Merton()].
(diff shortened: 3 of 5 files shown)


R/calc_survival_probability_merton.R
```diff
@@ -2,8 +2,8 @@
 #'
 #' Function calculates survival probability for a maturity based on a structural
 #' Merton model.
-#' For details on implementation please compare [CreditRisk::Merton()].
-#' Unlike [CreditRisk::Merton()] this implementation:
+#' For details on implementation please compare `CreditRisk::Merton()`.
+#' Unlike `CreditRisk::Merton()` this implementation:
```
maurolepore, line 6: We don't use the package CreditRisk. If we `[link it]`, R CMD check will expect us to Import: or Suggest it.

R/imports.R
```diff
@@ -46,6 +46,7 @@ globalVariables(
     "net_profit_margin",
     "pacta_sectors_not_analysed",
     "pd",
+    "plan_carsten",
```
maurolepore, line 49: I may have missed this one before, or it was added in the last few days.

DESCRIPTION
```diff
@@ -23,7 +23,7 @@ License: file LICENSE
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
```
maurolepore, line 26: Happens automatically when I redocument with updated packages. If you're behind you may run:

```r
update.packages(ask = FALSE)
```

.Rbuildignore
```diff
@@ -24,3 +24,4 @@
 ^wip_data_qa\.R$
 ^\.github$
 ^\.Renviron$
+^vignettes/articles$
```
maurolepore, line 27: This is in preparation for another PR.

MirijaHa (approved):

PR #136, opened 2021-11-01
New article template

maurolepore:
@jacobvjk please edit the file vignettes/articles/template.Rmd and add whatever content you want. The name of the file is generic; when you know a better one let me know.

You can keep pushing commits to this PR.


jacobvjk:
@maurolepore since this format is a bit different to the plain vignette format, my question is: should I still set up separate files for separate vignettes? Or do they all go into one file here?

maurolepore:
How you organize the information is up to you. Each file will generate a separate article. For example, see the ["Articles"](https://dplyr.tidyverse.org/) dropdown menu of dplyr, and compare it with the contents of [vignettes/](https://github.com/tidyverse/dplyr/tree/main/vignettes).

Note the yaml header of those articles has the field ["vignette:"](vignette: >) which we don't have. It's fine. The vignette field makes the article available in R with something like `vignette("dplyr", package = "dplyr")` -- which is nice but adds a more work on our end and make the package heavier to distribute. For now I suggest the lither weight article that is not a formal vignette but is still available on the package website under the section "Articles".

maurolepore:
I see articles already live. Closing this template as it serve no purpose.

PR #138, opened 2021-11-01
128 export advertised objects

### LINKED ISSUE #128: Export advertised objects
Do we need to export `lgd_senior_claims_range_lookup` (and other `*lookup` objects)?

In `run_stress_test_bonds()` I see:

> `lgd_senior_claims` ... for accepted value range check `lgd_senior_claims_range_lookup`.

But `lgd_senior_claims_range_lookup` is unavailable to users:

``` r
lgd_senior_claims_range_lookup
#> Error in eval(expr, envir, enclos): object 'lgd_senior_claims_range_lookup' not found
```

Only developers can see it.

``` r
devtools::load_all()
lgd_senior_claims_range_lookup
#> [1] 0.3 0.6
```

<sup>Created on 2021-10-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore:
Export and document _lookup objects
   
Closes #128
 
https://2degreesinvesting.github.io/r2dii.climate.stress.test/reference/lookup.html


R/lookup.R
```diff
@@ -2,43 +2,133 @@
 # practical purpose.
 investor_name_placeholder <- "Meta Investor"
 
+#' Help lookup values
```
MirijaHa, line 5: I am wondering if we should not better restrict the export to vars that are advertised in the documentation of run_stress_test, e.g. lgd_senior_claims_range_lookup and keep the remaining lookups as internal use only, as they should not be of relevance to the users.

maurolepore: Noted. TODO.

maurolepore:
Favour https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/143

R/lookup.R
maurolepore: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/143

PR #140, opened 2021-11-02
Ado 1669 names check

MirijaHa:
This closes ADO 1669
Mostly cosmetic.
I justified doing it now. because the error message in case columns are missing will be more helpful.


jacobvjk (approved):
lgtm

PR #142, opened 2021-11-02
Ado 1915 input checks

MirijaHa:
Only sector exposures


R/calculate_aum.R
```diff
@@ -15,5 +15,10 @@ calculate_aum <- function(sector_exposures) {
       asset_portfolio_value = sum(valid_value_usd),
       .groups = "drop"
     )
+
+  if (aum$asset_portfolio_value <= 0) {
+    stop("Calculated implausible asset under management.", call. = FALSE)
```
jacobvjk, line 20: maybe be more precise about the actual error: "Asset under management has negative value. This is not supported by the analysis." something like that. In principle one could imagine a portfolio with lots of short positions that add up to a negative overall value. It should not happen in our analyses, but it is not out of this world to imagine the situation

R/wrangle_and_check.R
```diff
@@ -15,9 +15,22 @@ wrangle_and_check_sector_exposures <- function(sector_exposures, asset_type) {
...
 
   if (!asset_type %in% c("Bonds", "Equity", "Loans")) {
-    stop("Invalid asset type")
+    stop("Invalid asset type", call. = FALSE )
   }
 
+  if (!is.logical(sector_exposures$valid_input) ||
+      !is.character(sector_exposures$financial_sector) ||
+      any(sector_exposures$valid_value_usd < 0)) {
+    stop("Detected implausible inputs on sector exposures", call. = FALSE)
```
jacobvjk, line 24: same as above. More detail me be helpful here

R/wrangle_and_check.R
```diff
@@ -15,9 +15,22 @@ wrangle_and_check_sector_exposures <- function(sector_exposures, asset_type) {
...
+  }
+
+  check_data_structure(
+    data = sector_exposures,
+    name_data = "sector exposures",
+    cuc_cols = "financial_sector",
```
jacobvjk, line 30: I think I disagree with this selection of CUC cols here. Made a comment in the data dictionary https://docs.google.com/spreadsheets/d/17VU49YwfIkv82dkFIg5EyZinp_A7EetlR9LrXP8eNlU/edit#gid=214413387

jacobvjk (commented):
see comments

R/wrangle_and_check.R
MirijaHa: you are so right, and I should have noticed by testing all asset_types, eq and cb would fail with the current cuc, fixed.

R/calculate_aum.R
MirijaHa: DONE

jacobvjk (approved):
lgtm

PR #144, opened 2021-11-02
Avoid warning about partial matching

maurolepore:
`devtools::test()` used to yield this warning which is now fixed:
    

    Warning (test-annual_pd_change_sector_shock_year.R:14:3): production trajectories function returns a ggplot object
    partial argument match of 'shock_year' to 'shock_year_filter'

### DIFF
--- a/tests/testthat/test-annual_pd_change_sector_shock_year.R
+++ b/tests/testthat/test-annual_pd_change_sector_shock_year.R
@@ -13,7 +13,7 @@ test_that("production trajectories function returns a ggplot object", {
 
   test_plot <- annual_pd_change_sector_shock_year(
     data = test_input_graph,
-    shock_year = test_shock_year,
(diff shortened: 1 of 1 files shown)


MirijaHa (approved):
May I ask how you spotted this warning? Neither when running devtools::test() not devtools::check() I can repro it. I use the most recent version of testthat (3.1.0) and did no funny business on customizing warning options or alike?

maurolepore:
Probabbly because my .Rprofile includes this:

```r
options(
  warnPartialMatchArgs = TRUE,
  warnPartialMatchDollar = TRUE,
  warnPartialMatchAttr = TRUE
)
```

https://github.com/maurolepore/dotfiles/blob/ead3aa1c42d019a8619ad0af471c21da24f39087/Rprofile#L29

PR #146, opened 2021-11-02
2587 Use gh action check full

maurolepore:
Relates to [AB#2587](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/2587)


maurolepore:
Pretty cool the package works with R back to 3 old releases! And I suspect we might be able to support one older release (R 3.5) if we rebuild the data/ via `usethis::use_data()`.

MirijaHa:
That is amazing to hear, thanks for checking :).
@jacobvjk  now you can also state the minimum version  (3.5.3) in the sys requirements in the  user facing manual.
3.5.3 is from march 2019 so > 2.5 years old, so pragmatically I would say it is not worth investing time in ensuring compatibility to even older versions, or what do you think @maurolepore?

maurolepore:
> ... pragmatically I would say it is not worth investing time in ensuring compatibility to even older versions, or what do you think @maurolepore?

I think supporting oldrel-3 (R 3.5.3 as of now) is good enough. For reference the tidyverse team supports oldrel-4 so not too far. And now that I read the warning more carefully I see the problem is not in data/ but in tests/testthat/tests_data/. So I'm not worried about users getting corrupt data. But I would spend a few minutes in re-reading those two datasets and re-saving them to see if the warning goes away. 


```
WARNING: Added dependency on R >= 3.5.0 because serialized objects in
     serialize/load version 3 cannot be read in older versions of R.
     File(s) containing such objects:
       ‘r2dii.climate.stress.test/tests/testthat/test_data/test_data-write-results-pf.rds’
       ‘r2dii.climate.stress.test/tests/testthat/test_data/test_interpolate_auto.rds’
```

Yet this PR seems already useful so I edited DESCRIPTION and the workflow so it can be merged right now. That way my short exploration doesn't block this PR.

.github/workflows/check-full.yaml
```diff
@@ -19,10 +23,17 @@ jobs:
...
+
           - {os: windows-latest, r: 'release'}
-          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-latest,   r: 'release'}
-          - {os: ubuntu-latest,   r: 'oldrel-1'}
+          # Use 3.6 to trigger usage of RTools35
+          - {os: windows-latest, r: '3.6'}
+
+          # Use older ubuntu to maximise backward compatibility
+          - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
```
MirijaHa, line 32: why did os changed fron ubuntu-latest to ubuntu-<version>?

maurolepore: Not sure about that -- I just run `usethis::use_github_action("check-full")` to get the latest version of that workflow (from r-lib/actions).

maurolepore:
I see some checks failing with

```
WARNING
  Dependence on R version '3.5.3' not with patchlevel 0
```

So I moved it back to a draft and will explore what's going on.

maurolepore:
I should use R > 3.5 for now

https://stackoverflow.com/questions/48433412/cran-check-warning-dependence-on-r-version-3-4-3-not-with-patchlevel-0﻿

.github/workflows/check-full.yaml
maurolepore: @MirijaHa please see if this answers your question and if you approve we can merge and get it out of the way.

MirijaHa: LGTM now, will approve.
Just a small future headache, as ubuntu version is now hardcoded we should make  even more sure to regularly update the gh actions workflow. Obviously one could simply set a monthly reminder to update it, but I wonder if you use a more informed system to make sure your workflows do not outdate?

MirijaHa (approved):
splendid

.github/workflows/check-full.yaml
maurolepore: Yeah, maintaining the workflows is indeed a quite time consuming task. I guess we could automate the process to run `usethis::use_github_actions("check-full")`, here and maybe all CRAN packages where we make an effort to keep the package compliant with that that workflow expects.

maurolepore: Maybe we could trigger the update only when a commit touches r-lib/actions/examples/check-full.yaml

PR #151, opened 2021-11-03
Use .data$ to avoid NOTE: Undefined global variables

maurolepore:
This PR avoids a NOTE we recently got rid of.

I could have added the variable to `globalVariables()` but I note the variable was introduced very recently, so maybe best to just tackle the underlying issue.

```
git log -S 'pull(financial_sector)'
 ...
Date:   Wed Nov 3 10:36:42 2021 +0100
```

I look forward to getting rid of all R CMD check issues so it's easier to notice when we inevitably introduce a new R CMD check issue

A painful reminder we still have many of these:

```
set_project_parameters: no visible binding for '<<-' assignment to
    ‘inc_stresstest’
```


MirijaHa (approved):
Jacob want to do a .data and .env rampage soon, to fix the issue for all :).
The other note comes from stuff for webtool, so we will be rid of it as soon as we split repos :)

PR #153, opened 2021-11-03
Enumarate valid inputs to asset type

maurolepore:
This PR polishes the interface of `run_stress_test()` to more meet a design patter: [Enumerate possible options](https://design.tidyverse.org/def-enum.html).

> If the possible values of an argument are a small set of strings, set the default argument to the set of possible values, and then use `match.arg()` or `rlang::arg_match()` in the function body. This convention advertises to the user what the possible values, and makes it easy to generate an informative error message for inappropriate inputs.

This prepares the interface to build the sensitivity analysis feature on top of it. Along the way I add:

* a `local_*()` function that allows skipping expensive computations in tests. 
* some tests including a snapshot of the warning the user will see if they don't provide an explicit `assert_type`.
* A use of `warn(c("Title", x = "Probleml", i = "Hint"))` for what it is but also as an example for how to produce nice condition messages (with a structure that helps meet the tidyverse style guide for such messages).

--

Note I suppress a ton of cluttering output and a warning. I flag FIXME.


.Rbuildignore
```diff
@@ -17,3 +17,4 @@
 ^\.github$
 ^\.Renviron$
 ^vignettes/articles$
+^inst/ns/model_parameters/\.gitkeep\.txt$
```
maurolepore: Not sure how this got here.

R/run_stress_test.R
```diff
@@ -26,7 +26,7 @@
 #'   excluded_companies.csv shall be excluded.
 #' @return NULL
 #' @export
-run_stress_test <- function(asset_type,
+run_stress_test <- function(asset_type = sort(asset_types_lookup),
```
maurolepore, line 29: This could be explicitely `c("bonds", "equity", "loans")`, but hiding those types behind an object seems a bit more maintenable.

R/run_stress_test.R
```diff
@@ -36,6 +36,58 @@ run_stress_test <- function(asset_type,
...
+  warn_if_asset_type_is_longer_than_1(asset_type)
+  matched_asset_type <- match.arg(asset_type)
+
+  # Allow testing anything before the long computation that follows
+  if (is_dev_env()) {
+    return(invisible(matched_asset_type))
+  }
+
+  # FIXME: Avoid clutter from printed "messages" that the user can't suppress
+  x <- capture.output(
```
maurolepore: It'd be nice to hear why we have so much printed output and confirm it's okay to suppress.

R/run_stress_test.R
```diff
@@ -36,6 +36,58 @@ run_stress_test <- function(asset_type,
                             shock_year = 2030,
                             term = 2,
                             company_exclusion = TRUE) {
+  warn_if_asset_type_is_longer_than_1(asset_type)
+  matched_asset_type <- match.arg(asset_type)
+
+  # Allow testing anything before the long computation that follows
+  if (is_dev_env()) {
+    return(invisible(matched_asset_type))
+  }
```
maurolepore, line 45: You may use `is_dev_env()` to do just about anything, but here I use it to return early so I can test the assertion above without running the more expensive computation below. See tests/

R/run_stress_test.R
```diff
@@ -36,6 +36,58 @@ run_stress_test <- function(asset_type,
...
+      risk_free_rate = risk_free_rate,
+      discount_rate = discount_rate,
+      div_netprofit_prop_coef = div_netprofit_prop_coef,
+      shock_year = shock_year,
+      term = term,
+      company_exclusion = company_exclusion
+    )
+  )
+
+  invisible(asset_type)
```
maurolepore, line 63: Maybe a more useful output would be the input data -- which suggests that maybe the first argument should be `data` and take an R object (e.g. a list of dataframes, or a nested dataframe) or a path to a folder with the expected input data. For backward compatibility default would be `data = Sys.getenv("ST_SUCH_AND_SUCH")`.

R/run_stress_test.R
```diff
@@ -36,6 +36,58 @@ run_stress_test <- function(asset_type,
...
+
+warn_if_asset_type_is_longer_than_1 <- function(asset_type) {
+  if (length(asset_type) > 1L) {
+    warn(c(
+      "`asset_type` should be of length 1.",
+      x = glue("It is of length {length(asset_type)}: {toString(asset_type)}."),
+      i = glue("Using default: {asset_type[[1]]}."),
+      i = glue(
+        "Suppress this warning e.g. with: `run_stress_test(asset_type = 'bonds')`."
+      )
```
maurolepore, line 74: The `message` argument to `rlang::abort()` and friends is super cool!
https://rlang.r-lib.org/reference/abort.html#arguments

R/utils-test.R
```diff
@@ -0,0 +1,23 @@
...
+#' * `is_dev_env()`: Returns the value of the option (`TRUE` or `FALSE`).
+#'
+#' @examples
+#' local({
+#'   local_dev_env()
+#'   is_dev_env()
+#' })
+#' is_dev_env()
+#' @noRd
+local_dev_env <- function(envir = parent.frame()) {
```
maurolepore, line 15: https://testthat.r-lib.org/articles/test-fixtures.html#local-helpers

DESCRIPTION
```diff
@@ -25,6 +25,7 @@ Imports:
     fs,
     ggplot2,
     glue,
+    here,
```
maurolepore, line 28: Not sure why `skip_on_ci()` and `skip_if_not_installed("here")` did not do the trick, and the checks were still expecting `here::here()` to be imported. For now I just added the dependency. Before a CRAN release I'd like to try prune as many packages as possible anyway.

R/run_stress_test.R
MirijaHa: We have the printed output because Jacob and me need it as long as we are still actively developing the model. Just today it helped me find another bug and I will even have to add more clutter for the next days.
Next week I will make sure to discriminate between output that is necessary for the user and may not be suppressed and outputs that we need only for development purpose which I will surpress when run from user mode. Till then please do not supress output by default.

R/run_stress_test.R
MirijaHa: - I put a general check that all input args must be of length 1 here https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/52b74a86abcfa084c269ee86eb1e8d852b148c34/R/validate.R#L18, just making sure your test extends this?
- asset type being of length >1 would have to result in an error, not a warning.
- I wonder if this function should be specific to asset_type. A function warn_if_arg_is_longer_than_1 could be recycled for other inputs that need to be of length 1.

R/run_stress_test.R
MirijaHa: Sounds like this will be helpful for e.g. for differntiating verbosity of outputs between developer and user mode.
However here it seems a bit intrusive as it does not serve stress testing functionality, is it here for proof of concept and will be removed shortly?

R/run_stress_test.R
```diff
@@ -36,6 +36,55 @@ run_stress_test <- function(asset_type,
                             shock_year = 2030,
                             term = 2,
                             company_exclusion = TRUE) {
+  warn_if_asset_type_is_longer_than_1(asset_type)
```
MirijaHa, line 39: fallback to default is not a pragmatically useful option for asset type, providing an asset_type of a length > 1 has to result in an error.

R/run_stress_test.R
```diff
@@ -26,7 +26,7 @@
 #'   excluded_companies.csv shall be excluded.
 #' @return NULL
 #' @export
-run_stress_test <- function(asset_type,
+run_stress_test <- function(asset_type = c("bonds", "equity", "loans"),
```
MirijaHa, line 29: while nicer from a design perspective I wonder if this is actually helpful from a user perspective. I see the danger that users do no understand that they have to explicitly select an input type, meaning their first interaction with the code would be a confusing error/warning, whereas if no default is provided it is obvious for the user what to do

R/run_stress_test.R
MirijaHa: I would not see a benefit in returning anything other than  NULL/invisible(asset_type) here.

R/run_stress_test.R
```diff
@@ -36,6 +36,55 @@ run_stress_test <- function(asset_type,
...
+        "Suppress this warning e.g. with: `run_stress_test(asset_type = 'bonds')`."
+      )
+    ))
+  }
+
+  invisible(asset_type)
+}
+
+run_stress_test_impl <- function(asset_type,
+                                 lgd_senior_claims = 0.45,
```
MirijaHa, line 79: I would suggest removeing default here, since they are passed on from the calling function

tests/testthat/test-run_stress_test.R
```diff
@@ -0,0 +1,23 @@
+test_that("is sensitive to `asset_type`, and defaults to bonds", {
```
MirijaHa, line 1: pls rm default to bonds, s.a.

tests/testthat/test-run_stress_test.R
```diff
@@ -0,0 +1,23 @@
...
+  local_dev_env()
+  expect_snapshot(out <- run_stress_test())
+  expect_equal(out, "bonds")
+})
+
+test_that("returns the first argument invisibly", {
+  skip_on_ci()
+  skip_on_cran()
+
+  readRenviron(here::here(".Renviron"))
```
MirijaHa, line 20: 1. can u please elaborate why readEnviron needs to be put here? it depends on using an .renviron on project level, which neither @jacobvjk  nor me currently do

tests/testthat/test-run_stress_test.R
```diff
@@ -0,0 +1,23 @@
...
+  expect_invisible(out <- purrr::quietly(run_stress_test)("bonds"))
```
MirijaHa, line 21: I am not entirely comfortable with having unit tests trigger run_stress_test. While of course it is beneficial to have it as an integration test:

- it would force us to set ST_PROJECT_FOLDER to a bonds compatible project every time we run devtools::test/check which would create overhead
- we had agreed in the team, that outputs in the ST_TESTING projects are always referring to results created from master. If running devtools::check on a dev branch we will mess with that. Strictly speaking this is not a problem, since ST_TESTING is supposed to be an experimental dev folder, but we would have to communicate this change in convention to the team.
- As a compromise: How about setting it in a way, that this test is skipped by default and needs to be explicitly activated?

R/run_stress_test.R
maurolepore: 1. Thanks, I saw that -- it's great you're asserting all inputs. But I think `asset_type` deserves special treatment. Compared to the other arguments, the possible values of this `asset_type` are small set of strings -- not a range of potentially infinite values. It fits really well in the description of the design pattern I link above. It's not a hard rule, of course, just a pattern. 
 
Here is one example in the wild:

``` r
head(usethis::use_blank_slate, 4)
#>                                           
#> 1 function (scope = c("user", "project")) 
#> 2 {                                       
#> 3     scope <- match.arg(scope)           
#> 4     if (scope == "user") {
```

2. If we use that pattern, then `match.arg()` is clever enough to throw informative errors.

``` r
f <- function(x = c("a", "b", "c")) {
  match.arg(x)
}

f(c("a", "b"))
#> Error in match.arg(x): 'arg' must be of length 1

f(c("d"))
#> Error in match.arg(x): 'arg' should be one of "a", "b", "c"
```

I'm also okay with using the first option silently, but I conservatively this time I chose to be more loud:

``` r
f <- function(x = c("a", "b", "c")) {
  if (length(x) > 1L) rlang::warn(glue::glue("Using default: {x[[1]]}."))
  match.arg(x)
}

f()
#> Warning: Using default: a.
#> [1] "a"
```

3. Yeah, as is this function is in an awckward middle gound. It's specific to `asset_type` but its name doesn't communicate that. Which way I take it depends on what we want:
    * a. Add "asset_type" to the name.
    * b. Make it general so we can use it elsewhere.
    
 The issue with b. is we might not need it. So I would start with a. and go with b. only if we later need it.

--

Of course, as every PR this is for discussion and I'm happy to keep the simpler approach we already have. Regardless how we handle `asset_type`, I think this PR already helped me think about how to test `run_stress_test()`. So happy to close and reuse the lessons I learn elsewhere.

MirijaHa (changes requested):

R/run_stress_test.R
maurolepore: 1. Thanks, I saw that -- it's great you're asserting all inputs. But I think `asset_type` deserves special treatment. Compared to the other arguments, the possible values of this `asset_type` are small set of strings -- not a range of potentially infinite values. It fits really well in the description of the design pattern I link above. It's not a hard rule, of course, just a pattern. 
 
Here is one example in the wild:

``` r
head(usethis::use_blank_slate, 4)
#>                                           
#> 1 function (scope = c("user", "project")) 
#> 2 {                                       
#> 3     scope <- match.arg(scope)           
#> 4     if (scope == "user") {
```

2. If we use that pattern, then `match.arg()` is clever enough to throw informative errors.

``` r
f <- function(x = c("a", "b", "c")) {
  match.arg(x)
}

f(c("a", "b"))
#> Error in match.arg(x): 'arg' must be of length 1

f(c("d"))
#> Error in match.arg(x): 'arg' should be one of "a", "b", "c"
```

I'm also okay with using the first option silently, but I conservatively this time I chose to be more loud:

``` r
f <- function(x = c("a", "b", "c")) {
  if (length(x) > 1L) rlang::warn(glue::glue("Using default: {x[[1]]}."))
  match.arg(x)
}

f()
#> Warning: Using default: a.
#> [1] "a"
```

3. Yeah, as is this function is in an awckward middle gound. It's specific to `asset_type` but its name doesn't communicate that. Which way I take it depends on what we want:
    * a. Add "asset_type" to the name.
    * b. Make it general so we can use it elsewhere.
    
 The issue with b. is we might not need it. So I would start with a. and go with b. only if we later need it.

--

Of course, as every PR this is for discussion and I'm happy to keep the simpler approach we already have. Regardless how we handle `asset_type`, I think this PR already helped me think about how to test `run_stress_test()`. So happy to close and reuse the lessons I learn elsewhere.

MirijaHa: 1 and 3 ok
2 _I'm also okay with using the first option silently, but I conservatively this time I chose to be more loud:_ Just to rule out we have no misunderstanding. The first option may not be chosen. In case asset_type is not of length 1 we need to throw an error.

MirijaHa: actually another thought on 1.
We should keep in mind that we write our code for another audience. Who choses to use usethis is probably a developer and used to reading function documentation, which we cannot assume from our users who will be more guided by intuition.
```
Also as opposed to head(usethis::use_blank_slate, 4)
#>                                           
#> 1 function (scope = c("user", "project")) 
#> 2 {                                       
#> 3     scope <- match.arg(scope)           
#> 4     if (scope == "user") {
```
in our case we have no useable default.

R/run_stress_test.R
maurolepore: OK, happy to leave `asset_type` alone. And if the distinction between asset types is so strong, then why not leave it as it was before -- with one interface per type. For example consider the specialized `purrr::map_*()` functions, like `map_int()`, `map_chr()`, etc. You could always use `vapply()` but the extra argument to specify the output tamplate is complex to the `map_*()` family feels friendlier -- particularly when coupled with auto-complete the user has a quick way to see what options there are (like with the enumeration approach) -- without having to read the docs.

MirijaHa: I think I do not fully understand your suggestion.
Are you suggesting going back to having 1 function per asset type? I would not do that because:

- Conceptually there is no real difference these days between the asset_types (historically they were but we cleaned that up over the last weeks). Remaining differences are mostly due to the fact that we need to handle inconsistencies between the input data. As these are dependent on PACTA results we cannot influence this and have to keep handling the inconsistencies.
- having 3 instead of functions triples copy paste work and testing needs on every change and also bears the danger of reintroducing inconsistencies 
Or are u suggesting providing 3 different interfaces that then call run_stress_test with arg "asset_type" set correctly? That Idea I would be very in favor of, but I do not see how it relates to the map family?

R/run_stress_test.R
maurolepore: Great you think of usability. With that in mind again I would think of 1 interface per option -- which reminds me of reading about affordances in The design of eveyday things. A quick google search throws me this:

> Affordance is the relationship between what something looks like and how it's used. For designers, it means that as soon as someone sees something, they have to know how to use it. For example, a mug has high affordance: it's easy to figure out intuitively how to use it.

R/run_stress_test.R
MirijaHa: ok, resolving

R/run_stress_test.R
MirijaHa: Great, as long as the interface is simply an interface to calling run_stress_test with the correct asset_type  as commented in more detail above let us go that way :)

MirijaHa: alternative, depening on how you plan to provide the interface for the sensitivity analysis, we could also place the distinguiation between the 3 types there?

tests/testthat/test-run_stress_test.R
maurolepore: RE:

> I would not see a benefit in returning anything other than NULL/invisible(asset_type) here.

Here I'm inspired by https://design.tidyverse.org/out-invisible.html

> If a function is called primarily for its side-effects, it should invisibly return a useful output. If there’s no obvious output, return the first argument. This makes it possible to use the function with in a pipeline.

``` r
# Good
f <- function(x) {
  writeLines("Nothing is always something", tempfile())
  invisible(x)
}

1 |> f() |> sum(2)
#> [1] 3

# Bad
f <- function(x) {
  writeLines("Nothing is always something", tempfile())
}

1 |> f() |> sum(2)
#> [1] 2
```

tests/testthat/test-run_stress_test.R
maurolepore: It's  make the environmental variables available inside the test. The testing environment is independent and what works interactively may not work in a test. 

I didn't quite understand how you set the environmental variables. I could try re-read that but easiest to test is to not need environmental variables.

tests/testthat/test-run_stress_test.R
MirijaHa: here again I would want to be pragmatic and consider the purpose of the function. The sole purpose of the function is its sideeffect, i.e. wrapping around a pipeline of data import, some calculation, and export. The nature of the function make it not useful to potentially include it in a pipeline, thus I would opt for a NULL/invisible() return here.

tests/testthat/test-run_stress_test.R
MirijaHa: i simply use a global .renv, which is thus not readable via here::here.

Omitting env vars would force us to add two input args to determine the data sources locations and even then we would have to interact with the file system which I would prefer the test not to do. I am afraid with this limiting factor it is not possible to make a clean test unless we:

- abstract away bits of the code that interact, and have to interact, with the file system and test the remaining code in integration, using synthetic input data. 
- or setup a temp env with all needed inputs, again using synthetic input data
- While that is among the first things on the prio list after the deadline it is not realistic to implement before, thus I would suggest to either remove this test or disable it per default.

R/utils-test.R
```diff
@@ -21,3 +21,7 @@ local_dev_env <- function(envir = parent.frame()) {
 is_dev_env <- function() {
   getOption("r2dii.climate.stress.dev_env") %||% FALSE
 }
+
+is_mauro <- function() {
```
MirijaHa, line 25: hahahha, love it :).

tests/testthat/test-run_stress_test.R
maurolepore: 1. That's what my project-specific .Renviron does. Note it points to a directory that nobody else has, so there is no way to mess production directories.

``` r
writeLines(readLines(".Renviron"))
# Where to save project agnostic files?
# For example, see 2DII's Dropbox at
# PortCheck/00_Data/07_AnalysisInputs/ST_INPUTS_MASTER
ST_DATA_PATH=/home/mauro/tmp/ST_INPUTS_MASTER

# Where to save project specific inputs and outputs?
# For example, for bonds (fastest) see 2DII's Dropbox at 
# PortCheck_v2/10_Projects/ST_TESTING_BONDS
ST_PROJECT_FOLDER=/home/mauro/tmp/ST_TESTING_BONDS
```

2. I don't understand this. But it seems very different to what the usual development workflow is. It seems people are using the package in developer-mode to do analyses they should be doing in user-mode. Because the package is now usable in user-mode, analysts should be able to create a project, install this package, call `library()` and start using it. They should not need the source code, and their project should have the .Renviron file pointing wherever they want the IO to live.

I think we should talk in a meeting about this. Because a small change in the workflows might bring a great increase in quality-of-work-life.

3. How about setting it in a way, that this test is skipped by default and needs to be explicitly activated?

Good idea! While I wait to better understand what's the source of the workflow problem, this test will run only in my computer :)

```r
  skip_if_not(is_mauro())
```

R/run_stress_test.R
maurolepore: > are u suggesting providing 3 different interfaces that then call run_stress_test with arg "asset_type" set correctly?


Yes -- at least to reconsider it. Internally it's always best remove as much duplication as possible. But facing the users it's best to provide whatever it's more user-friendly -- which in this case might be three functions of the same family rather than one function with an additional argument that takes three valid (asset) options.

The relationship between this and the `map_*()` family is that the `map_*()` family could also have been exposed as a single function with an additional argument, but instead it is exposed as three different functions that call the same implementation with three different arguments -- interlally.

``` r
purrr::map_chr
#> function (.x, .f, ...) 
#> {
#>     .f <- as_mapper(.f, ...)
#>     .Call(map_impl, environment(), ".x", ".f", "character")
#> }
#> <bytecode: 0x562907657e18>
#> <environment: namespace:purrr>

purrr::map_int
#> function (.x, .f, ...) 
#> {
#>     .f <- as_mapper(.f, ...)
#>     .Call(map_impl, environment(), ".x", ".f", "integer")
#> }
#> <bytecode: 0x562907055090>
#> <environment: namespace:purrr>

purrr::map_dbl
#> function (.x, .f, ...) 
#> {
#>     .f <- as_mapper(.f, ...)
#>     .Call(map_impl, environment(), ".x", ".f", "double")
#> }
#> <bytecode: 0x5629069c1f48>
#> <environment: namespace:purrr>
```

<sup>Created on 2021-11-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore:
Moving to a draft. I'll prepare a separate PR based on lessons learned from this discussion.

R/run_stress_test.R
maurolepore: Oh, right!

maurolepore:
I'll close. I can revisit these lessons anyway.

PR #154, opened 2021-11-03
2493 fin data wrangling to prep

jacobvjk:
closes ADO 2493: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2493

This PR:
- reads in prewarngled financial data inputs for profit margins and credit risk inputs from one central file that works for all asset types, rather than three separate ones.
- It moves the last prewrangling steps to the data repo

NOTE:
- this will lead to result changes in all asset types and output files. We previously erroneously discarded some company/tech combinations, sometimes entire companies. Since including these via https://github.com/2DegreesInvesting/r2dii.stress.test.data/pull/39 will lead to an overall slightly different composition of considered companies and production, the aggregated values in all files will change. The changes are small, as only a handful of companies is affected and they do not inverse any trends previously seen


R/read_company_data.R
```diff
@@ -1,42 +1,25 @@
-#' Read in company finacial data processed from eikon exports and AR master data
-#' that contain information on multiple credit risk inputs and company
-#' production plans, aggregated to the ticker/technology/year level for
-#' `asset_type` bonds and aggregated to the company/technology/year level for
-#' `asset_type` equity and aggregated to the loan_taker/technology/year level
-#' for `àsset_type` loans
+#' Read in company financial data processed from eikon exports and AR master data
```
MirijaHa, line 1: please rename file (git mv old new if u want to be facny)

R/wrangle_and_check.R
```diff
@@ -113,40 +113,27 @@ wrangle_and_check_pacta_results <- function(pacta_results, start_year, time_hori
...
+    # ADO 2493 - if asset_type not bond, ticker not required. Use distinct_all
+    # to remove duplicates from remaining CUC columns since financial data is
+    # always equal for these columns
+    financial_data <- financial_data %>%
+      dplyr::select(
+        .data$company_name, .data$company_id, .data$pd, .data$net_profit_margin,
+        .data$debt_equity_ratio, .data$volatility
+      ) %>%
+      dplyr::distinct_all()
+  }
```
MirijaHa, line 136: small logic question. Is it possible for 1 company to have several rows under different tickers? Jut asking to understand if the distinct::all makes sense.

MirijaHa (commented):
1 tiny remark and 1 question, otherwise lgtm, the harmonisation of input files is very welcome

R/wrangle_and_check.R
jacobvjk: will figure that out in the distinct_all() follow up task in this sprint :)

MirijaHa: thats the spirit

R/read_company_data.R
jacobvjk: done

MirijaHa (approved):
miraculous

PR #156, opened 2021-11-04
use valid license

MirijaHa:
From the CRAN readiness blockers that  @maurolepore  identified I figured invalid license is the most relevant aspect to fix also before GH ready release. I based the license on https://github.com/2DegreesInvesting/r2dii.plot/blob/develop/LICENSE
This should be formally correct, since there was a more restrictive license before though I would also like to have @jacobvjk approval before merging.


jacobvjk (approved):
lgtm

PR #157, opened 2021-11-04
Logging

MirijaHa:
This PR closes https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2394  and https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2550

- Logs of progress are a bit more verbose and formatted nicer as a courtesy to the users  (@jacobvjk  please check if the logs are actually saying the truth though)

- Conditions are signalled to the user based on their relevance:
1.  standard log messages with cat line
2. informative/analysis dependent information with an indented cat line
3. aspects that are possibly explain worthy (e.g. losing rows that are not fit to merton model)and optionally require action as warning
4. anything that requires action leads to an error.

- per default verbose diagnostic logs are switched of nowf. inspired by @maurolepore  they can be enabled by running verbose_log_env() before running the analysis. Currently this mode is not exported, if u think that a user might want to use it too I can change this @jacobvjk . User output looks like this now:

```
run_stress_test("equity")
 -- Validating input arguments. 
 -- Configuring analysis settings. 
 -- Importing and preparing input data from designated input path. 
 -- Calculating market risk.
 Warning: Removed 273 rows because negative production compensation targets were set in late and sudden.
                    Negative absolute production is impossible
      >> When joining annual profits on financial data on columns company_name ald_sector technology dropped 2310 rows from annual profits 
      >> When joining plan carsten on financial data on columns company_name ald_sector technology dropped 115 rows from plan carsten 
 -- Calculating credit risk. 
 -- Exporting results to designated output path. 
 -- Exported results to designated output path.

```

- change checks pass (tested for equity)


jacobvjk:
I wonder if it may be helpful to additionally add the relative amount of rows dropped at each stage. To indicate the severity.

jacobvjk (approved):
lgtm

MirijaHa:
> I wonder if it may be helpful to additionally add the relative amount of rows dropped at each stage. To indicate the severity.

happy to, will do in a follow up. I also wonder if to add logs for missing created here https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2180 I cannot give detailed information but what happens, but at least we would keep track of the affected numbers.

maurolepore (approved):
I review this quickly as I think @jacobvjk is the best person to judge it. Beyond my nit suggestions here, I have a couple of general thoughts:

1. Logs might be better handled via:

    a. a dedicated logging package (there may be something on CRAN); or
    b. written to a file for later inspection;
    c. printed with functoins smarter than `cat()`, e.g. `message()`,  `rlang::info()`, `usethis::ui_info()`, or something from the 'cli' package. These functions allow suppressing, and/or nicer formatting.
    
2. If the user might be interested in the logs, maybe it's best to expose that feature via an argument rather than an option.

PR #158, opened 2021-11-04
Fix tiny typo and add LICENSE.md via use_mit_license()

maurolepore:
This PR does some work automatically via `usethis::use_mit_license()`, just to ensure we're not forgetting any important detail. Indeed if found a lil typo and a missing file.


MirijaHa (approved):
good catch

PR #160, opened 2021-11-04
2647 financial data from csv

jacobvjk:
closes ADO 2647: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2647

all change checks passed


MirijaHa (commented):
- pls merge with master before approval
- we might want to simply drop those few rows that are not ASCII encoded as they should be useless any, or we could log their share as info to the user, but that would be entering unknown terrain. A conventinet way to detect non ASCII could be 
```
 dplyr::mutate(enc = 
stringi::stri_enc_mark(company_name))  %>% 
  dplyr::filter(enc != "ASCII") 
```

DESCRIPTION
```diff
@@ -32,6 +32,7 @@ Imports:
     r2dii.plot,
     readr,
     readxl,
+    stringi,
```
MirijaHa, line 35: please remove this unneeded dep, will approve then

MirijaHa (commented):
1 last comemnt

DESCRIPTION
jacobvjk: done

MirijaHa (approved):

PR #162, opened 2021-11-05
Fin data validataion

MirijaHa:
This PR adds extended input validation for financial_data.

Some context info and questions for dear @maurolepore 
- financial_data is the only dataset that users might create manually. Therefore the checks are more pendantic than for other input data which either we provide or comes from PACTA calculation. Also users might want to use the function check_financial_data to check their data before running stresstest, and might have to understand it in case it detects problems, which is not unlikely. Thus I exported it and wrote it in a way that should be easily understandable to someone without R/tidverse specific knowledge, i.e. I used no pipes or alike.
- preach wine drink water style I implemented this to cater to a bunch of specs that Jacob and me briefly discussed but did not note down as requirements in ADO, so dont feel like you have to review the content of the PR, however I would ask for some advice on including examples in function doc (compare check_financial_data).

1. would you recommend wrapping examples in dontrun? This example is not timeconsuming and is on an exported function so I did not see the need, but maybe you have another opinion?
2. do you know of a convenient way to style examples? Using styler add-in did not work for me, so I styled it manual which is probably not the wisest approach :)


R/wrangle_and_check.R
```diff
@@ -138,8 +167,27 @@ wrangle_financial_data <- function(financial_data, asset_type) {
         .data$debt_equity_ratio, .data$volatility
       ) %>%
       dplyr::distinct_all()
+  } else {
+    financial_data <- financial_data %>% dplyr::filter(!is.na(corporate_bond_ticker))
```
jacobvjk, line 171: .data$corporate_bond_ticker

R/wrangle_and_check.R
```diff
@@ -193,3 +240,44 @@ fill_annual_profit_cols <- function(annual_profits) {
...
+      stop("Implausibe value(s) < 0 for debt_equity_ratio detected. Please check.")
+    }
+  } else {
+    if (any(financial_data$debt_equity_ratio <= 0)) {
+      stop("Implausibe value(s) <= 0 for debt_equity_ratio detected. Please check.")
+    }
+  }
+
+  if (any(financial_data$volatility < 0)) {
+    stop("Implausibe value(s) < 0 or >= 1 for volatility detected. Please check.")
```
jacobvjk, line 281: the sentence doe not match the condition

R/wrangle_and_check.R
```diff
@@ -193,3 +240,44 @@ fill_annual_profit_cols <- function(annual_profits) {
...
+#'
+#' @inheritParams check_financial_data
+#'
+#' @return NULL
+check_valid_financial_data_values <- function(financial_data, asset_type) {
+  n_unique_ids <- length(unique(financial_data$company_id))
+  n_unique_names <- length(unique(financial_data$company_name))
+  n_unique_id_names_combinations <- length(unique(paste(financial_data$company_id, financial_data$company_name)))
+
+  if (length(unique(c(n_unique_ids, n_unique_names, n_unique_id_names_combinations))) > 1) {
```
jacobvjk, line 258: This implies that users will have to add the correct `company_id` for loans. That is only realistic if they get the company id from their pacta run, how else would they know the id? However P4B does not provide the `company_id` and we return `NA_real_` for all companies when wrangling the P4B results to P4I format. In my understanding this test would fail in that case

jacobvjk (commented):
see comment on `company_name` to `company_id` mapping

R/wrangle_and_check.R
MirijaHa: done

R/wrangle_and_check.R
MirijaHa: done

maurolepore:
@MirijaHa,

### 1. Examples

For a GitHub release I would focus mainly on making examples run so the user can see the output on the website -- without having to install or run anything ([example](https://2degreesinvesting.github.io/r2dii.climate.stress.test/reference/dropbox_exists.html#examples)). That might be a lil tricky with all the setup this package requires, but I'm happy to try. When the time comes it would very much help export only the main user-facing functions and nothing else. Then I'll know where to focus my effort.

For a CRAN release we'll need to be very careful. Recently CRAN got very picky about including examples, and the use of dontrun. A great source of wisdom is [this repo](https://github.com/DavisVaughan/extrachecks) -- which includes some tips specifically about examples. (That repo is linked in the release issue.).

A [hot new feature](https://roxygen2.r-lib.org/news/index.html#roxygen2-7-1-2-2021-09-08) that may help and I'm keen to try is `@examplesIf`. Maybe you'll try it first :-)
 
### 2. Styling examples

I apply the tidyverse style to roxygen examples with the styler addin plus some manual tweaks ([4' demo](https://youtu.be/0oXa112SUcM)).

jacobvjk (approved):
lgtm

PR #163, opened 2021-11-05
adjust handling of start year in loans stress test

jacobvjk:
closes ADO 2668: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2668

picks the right start year and files for loans preparation


R/lookup.R
```diff
@@ -139,7 +139,7 @@ p4b_scenarios_lookup <- c("target_b2ds", "target_cps", "target_rts",
                           "target_sds")
 
 # allowed input years p4b production data file
-p4b_production_data_years_lookup <- c(2020, 2021)
+p4b_production_data_years_lookup <- c(2019, 2020)
```
MirijaHa, line 142: just to go sure cfg$AnalysisPeriod$Years.Startyear, which used to determin the start_year was 2021 in the ST_TESTING projects before we manually corrected it. Just making sure we do not overlook anything in this regard?

R/prep_and_calc_loan_book.R
```diff
@@ -16,7 +19,7 @@ run_prep_calculation_loans <- function(credit_type = "outstanding") {
 
   cfg <- config::get(file = file.path(get_st_data_path("ST_PROJECT_FOLDER"), "inputs", "AnalysisParameters.yml"))
 
-  start_year <- cfg$AnalysisPeriod$Years.Startyear
+  start_year <- start_year
```
MirijaHa, line 22: u can remove this line now

R/prep_and_calc_loan_book.R
```diff
@@ -116,7 +119,7 @@ run_prep_calculation_loans <- function(credit_type = "outstanding") {
         ald_timestamp = "c"
       )
     )
-  } else if (start_year == 2021) {
+  } else if (start_year == 2020) {
```
MirijaHa, line 122: I do not entirely understand what is going on but:

- this will change the ald data we use. This means our pacta_results and sector_exposures as used in the main workflow will change also, are any structural changes possible?
- is it realistic we have to support 2019 in general?
- did you verify all other aspects that depend on start year where changed accordinglt?

R/lookup.R
jacobvjk: for loans, it is currently not possible to have start year 2021. it's always either 2019 or 2020, based on the scenario data start years that they have. I previously assumed identical handling as in P4I, but this is not the case. so yes, for now this is correct

R/prep_and_calc_loan_book.R
jacobvjk: I am ware this will change it, which again is due to my previous misconception that years are handled equally i P4I and P4B. I is correct now and will indeed change the values in the LOANS test case. It is a valid point that we may not have to support 2019 at all. I don't think anyone will be using it, but will check in with George

jacobvjk: only 2020 left now

R/prep_and_calc_loan_book.R
jacobvjk: done

R/lookup.R
jacobvjk: completely removed this for now (only 2020 supported hard coded for now)

MirijaHa (commented):
For peace of mind:

- how high are the deviations when using those pacta results vs the ones with the old workflos?

otherwise lgtm

jacobvjk:
cfg removed, change checks pass

MirijaHa (approved):

PR #165, opened 2021-11-05
rm non negative nad add raw total loan book size to overview

jacobvjk:
closes ADO 2690: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2690

This PR:
- removes negative loans of the select loan type in the preparation script
- gets the raw total loan size for the total value of the `overview_portfolio` file
- change checks passed

### DIFF
--- a/R/prep_and_calc_loan_book.R
+++ b/R/prep_and_calc_loan_book.R
@@ -103,6 +103,7 @@ run_prep_calculation_loans <- function(credit_type = "outstanding") {
   )
 
   #### Wrangle and prepare data-------------------------------------------------
+  # ADO 2690 - remove rows with negative loan values (not allowed in P4B)
   if (credit_type == "outstanding") {
     matched_non_negative <- matched %>%
       dplyr::filter(.data$loan_size_outstanding >= 0)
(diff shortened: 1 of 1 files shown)


jacobvjk:
change checks still open

R/prep_and_calc_loan_book.R
```diff
@@ -103,17 +103,14 @@ run_prep_calculation_loans <- function(credit_type = "outstanding") {
...
-        .data$loan_size_credit_limit < 0, 0, .data$loan_size_credit_limit
-      )
-    )
+  # ADO 2690 - remove rows with negative loan values (not allowed in P4B)
+  if (credit_type == "outstanding") {
+    matched_non_negative <- matched %>%
+      dplyr::filter(.data$loan_size_outstanding >= 0)
+  } else {
+    matched_non_negative <- matched %>%
+      dplyr::filter(.data$loan_size_credit_limit >= 0)
```
MirijaHa, line 112: just to make sure, since in old version you corrected negative values for both credit_types. Is it the case that the column of the not selected credit_type  is of no relevance?

R/prep_and_calc_loan_book.R
```diff
@@ -103,17 +103,14 @@ run_prep_calculation_loans <- function(credit_type = "outstanding") {
...
+  }
```
MirijaHa, line 113: I wonder if this dropping of rows should be related some user facing information (log message or warning), indicating which companies are affected?

MirijaHa (commented):
2 comments out of interest

R/prep_and_calc_loan_book.R
jacobvjk: yes, I think this is actually safer, because only one is of relevance for the further calculation. Thus, if for whatever reason, the other column is negative, this should not be removed

R/prep_and_calc_loan_book.R
jacobvjk: done

MirijaHa (approved):

PR #166, opened 2021-11-07
New functions for sensityvity analysis

maurolepore:
This PR implements new functions to perform a sensitivity analysis: [`st_bonds()`, `st_write_bonds()`, and friends.](https://2degreesinvesting.github.io/r2dii.climate.stress.test/reference/st_bonds.html).

[Reference](https://2degreesinvesting.github.io/r2dii.climate.stress.test/reference/index.html).


R/st_bonds.R
```diff
@@ -0,0 +1,194 @@
...
+  st_df(data, asset_type = "loans", ..., quiet = quiet)
+}
+
+#' @rdname st_bonds
+#' @export
+st_write_bonds <- function(data = st_data_paths(),
+                           destdir = tempdir(),
+                           ...,
+                           quiet = TRUE) {
+  st_write(data, asset_type = "bonds", destdir = destdir, ..., quiet = quiet)
```
maurolepore, line 83: It's silly I read and re-write the output, instead of moving it to `destdir`. But for now I'm worried only about the interface and to not touch existing code. Later we can refactor for better performance.

R/st_bonds.R
```diff
@@ -0,0 +1,194 @@
...
+  for (i in seq_along(args)) {
+    out <- exec(st, !!!args[[i]])
+    out <- unnest(out, st_result)
+    write_csv(out, file = path(destdir, glue("{names(args)[[i]]}.csv")))
+  }
+
+  invisible(data)
+}
+
+enlist_args <- function(data, asset_type, ..., quiet) {
```
maurolepore, line 126: I think this is the most interesting part of this PR: Creating a list with as many elements as values are passed to one and only one of the arguments that are forwarded to `run_stress_test()` (via `...`). Each element has those arguments to be forwarded, plus the arguments to `st()` -- which is a thin wrapper of `run_stress_test()` with a signature that is easier to work with.

R/st_bonds.R
```diff
@@ -0,0 +1,194 @@
...
+abort_if_no_argument_is_long <- function(data) {
+  if (identical(length(data), 0L)) {
+    abort(c(
+      "Must privide one argument of `run_stress_test()` with multiple values.",
+      i = "Do you need to use `run_stress_test()` instead?"
+    ))
+  }
+
+  invisible(data)
+}
```
maurolepore, line 150: This is pedantic and annoying but takes the least code. Later we could support a smarter version that knows how to solve this issue.

R/st_bonds.R
```diff
@@ -0,0 +1,194 @@
...
+
+abort_if_more_than_one_argument_is_long <- function(data) {
```
maurolepore, line 152: This is a requirement.

R/st_bonds.R
```diff
@@ -0,0 +1,196 @@
...
+  data %>%
+    separate(.data$name, into = c("arg_name", "arg_value"), sep = "___") %>%
+    unnest(cols = .data$st_result) %>%
+    relocate(.data$st_type, .data$st_name)
+}
+
+st <- function(data, asset_type, ..., quiet = TRUE) {
+  vec_assert(data, character(), size = 2L)
+  data <- vec_set_names(data, envvar_names())
+  local_envvar(data)
```
maurolepore, line 174: This sets the environment variables given the paths in `data`. The user does not need to know how to do that.

R/st_bonds.R
```diff
@@ -0,0 +1,196 @@
...
+
+  control_verbosity <- ifelse(quiet, utils::capture.output, identity)
+  invisible(control_verbosity(run_stress_test(asset_type = asset_type, ...)))
+
+  outputs_path() %>%
+    dir_ls() %>%
+    map(~ read_csv(.x, show_col_types = FALSE)) %>%
```
maurolepore, line 181: It's wasteful to write data (with `run_stress_test()`) then re-read it here. But for now I focus building a function that is easier to program with (inputs `data` and outputs a data frame), and that doesn't touch existing code.

 Later we may refactor to improve performance.

_pkgdown.yml
```diff
@@ -0,0 +1,11 @@
+url: ~
+
+reference:
+- title: Main functions
+  contents:
+  - has_concept("main functions")
+- title: Helper functions
+  contents:
+  - has_concept("helpers")
+  - lookup
```
maurolepore, line 10: The setup of the website seems to belong to a separate PR. I may do that then fix the merge conflicts.

tests/testthat/_snaps/st_bonds.md
```diff
@@ -0,0 +1,11 @@
...
+
+    Must privide one argument of `run_stress_test()` with multiple values.
+    i Do you need to use `run_stress_test()` instead?
+
+# with more than one long stress-test argument errors gracefully
+
+    Must provide no more than one argument with multiple values.
+    x Arguments with multiple values: lgd_senior_claims, terminal_value.
+    i Did you forget to pick only one?
+
```
maurolepore, line 11: These are the error messages the user will see. Please review and suggest the wording you prefer (ideally still following the tidyverse style guide for error messages).

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+#' @rdname st_bonds
+#' @export
+st_write_loans <- function(data = st_data_paths(),
+                           destdir = tempdir(),
+                           ...,
+                           quiet = TRUE) {
+  st_write(data, asset_type = "loans", destdir = destdir, ..., quiet = quiet)
+}
+
+st_df <- function(data, asset_type, ..., quiet = TRUE) {
```
MirijaHa, line 104: I would suggest to use a more descriptive function name

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+}
+
+restructure_st_df <- function(data) {
+  data %>%
+    separate(.data$name, into = c("arg_name", "arg_value"), sep = "___") %>%
+    unnest(cols = .data$st_result) %>%
+    relocate(.data$st_type, .data$st_name)
+}
+
+st <- function(data, asset_type, ..., quiet = TRUE) {
```
MirijaHa, line 171: please use more descriptive fucction name

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+      x = glue("Arguments with multiple values: {toString(names(data))}."),
+      i = "Did you forget to pick only one?"
+    ))
+  }
+
+  invisible(data)
+}
+
+restructure_st_df <- function(data) {
+  data %>%
```
MirijaHa, line 165: are the implicit returns in several functions intended?

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+#'
+#' @examplesIf r2dii.climate.stress.test:::is_registered_dev("/home/mauro")
+#' library(readr, warn.conflicts = FALSE)
+#'
+#' # Give the paths to your data explicitely
+#' data <- st_data_paths(
+#'   data = "/home/mauro/tmp/st/ST_INPUTS_MASTER",
+#'   project = "/home/mauro/tmp/st/ST_TESTING_BONDS"
+#' )
+#' out <- suppressWarnings(
```
MirijaHa, line 31: I assume out is all data combined in one df? I wonder if this output is not too overwhelming to users to be of use, or if it would, worst case, convey misleading information, as the data result sets are not meant to be considered together.

MirijaHa (commented):
This implementation is powerful in giving the user options and allowing the user to inspect data in places other than the input/output dirs. However this brings along more complicated use. Given our user base (people without R or dev experience who only use our code to generate the output csvs) I think we should cut down on power and make the use as simple as possible. I included a few suggestions along the way.

Three separate points though:

- I could not review the write functions. As far as I see they are not called from st_<asset_type> and I did not manage to get them running independently.
- when running st_<asset_type> using this command https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/f3fbd235e56b145937aae40085541270af760426/R/st_bonds.R#L39 I received the warning `One or more parsing issues`. The warning is thrown here https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/f3fbd235e56b145937aae40085541270af760426/R/st_bonds.R#L111 , I assume the root cause happens when calling https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/f3fbd235e56b145937aae40085541270af760426/R/st_bonds.R#L181 in the second run. I had this problem with reading in stress testing results before when it still included iterations in run_stress_test, will try to provide you with a more specific account of the problem. If we cannot easily fix this I would suggest changing run_stress_test so it returns the data instead of exporting them here https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/5a53846617d223eb9f6a4e4c8e8de3399ac5074d/R/run_stress_test.R#L192. Then we simply need to rowbind them an then call write_stress_test_result. As this saves us reformatting et cetera this might be the prefered solution anyway.
- easy nit:  name and st_result are undefined global

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+st <- function(data, asset_type, ..., quiet = TRUE) {
+  vec_assert(data, character(), size = 2L)
+  data <- vec_set_names(data, envvar_keys())
+  local_envvar(data)
+
+  control_verbosity <- ifelse(quiet, utils::capture.output, identity)
+  invisible(control_verbosity(run_stress_test(asset_type = asset_type, ...)))
+
+  outputs_path() %>%
+    dir_ls() %>%
```
MirijaHa, line 180: robustness: what happens in the, rather likely, case that users randomly drop files in this folder along the way?

R/st_bonds.R
jacobvjk: what would be a better option though? if the user passes 10 variations of a parameter, which I consider entirely plausible for a sensitivity analysis, and we save the results to separate files, the output directory will be cluttered and hard to navigate. What is more, the intuitive way to interpret results from a sensitivity analysis would be to visualise the variations in one graph, imo. One df seems like a good solution for that?

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+#' # Later you may read all results at once
+#' files <- list.files(path, full.names = TRUE)
+#' results <- read_csv(files, id = "arg_val", show_col_types = FALSE)
+#' results
+#'
+#' # The column `arg_val` helps you identify each run
+#' unique(
+#'   subset(results, grepl("term.*2", arg_val), select = 1:3)
+#' )
+st_bonds <- function(data = st_data_paths(), ..., quiet = TRUE) {
```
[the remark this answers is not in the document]
jacobvjk: agreed

R/st_bonds.R
```diff
@@ -0,0 +1,185 @@
...
+#'
+#' These functions performs the so called "sensitivity analysis" for the asset
+#' type in their name. They run [run_stress_test()] multiple times, iterating
+#' over multiple values of one and only one argument of [run_stress_test()].
+#'
+#' @param data Character vector of length 2. Paths to the directories
+#'   ST_INPUTS_MASTER and ST_TESTING_\{aset-type\}, e.g. ST_TESTING_BONDS,
+#'   respectively. These directories must have the files expected by the
+#'   stress-test project. The easiest way to create this object is with
+#'   [st_data_paths()].
```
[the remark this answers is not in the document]
jacobvjk: In general I believe users will appreciate not having to set env vars, but I also agree the interface for the required paths should be simple for users that are not used to R or any tech heavy work.
In that sense i like two separate input args for project_path and agnostic_path (do not use my naming here), and that the output path is simply derived from the input path (at least by default)

R/st_bonds.R
```diff
@@ -0,0 +1,192 @@
...
+
+#' @rdname st_bonds
+#' @export
+st_loans <- function(data = st_data_paths(), ..., quiet = TRUE) {
+  st_df(data, asset_type = "loans", ..., quiet = quiet)
+}
+
+#' @rdname st_bonds
+#' @export
+st_write_bonds <- function(data = st_data_paths(),
```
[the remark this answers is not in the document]
jacobvjk: Agreed, I do not see the added benefit of having a separate function for each asset type which then calls all the same functions regardless?

R/utils-tests.R
```diff
@@ -0,0 +1,23 @@
...
+#' @noRd
+is_registered_dev <- function(my_home = "/home/mauro") {
+  identical(path.expand("~"), my_home)
+}
+
+expect_no_output <- function(object, ...) {
+  testthat::expect_output(object, regexp = NA, ...)
+}
+
+expect_no_error <- function(object, ...) {
```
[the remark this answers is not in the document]
jacobvjk: noted, cool!

R/imports.R
```diff
@@ -1,4 +1,14 @@
-#' @importFrom rlang %||% abort warn .data .env
+#' @importFrom vctrs vec_assert vec_set_names
```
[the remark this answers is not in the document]
jacobvjk: I do not fully understand the alternatives that are being suggested

jacobvjk (commented):
Some first comments, mostly content/behaviour related

R/st_bonds.R
MirijaHa: The concrete problem here is that the out dataframe combines the 5 results types as far as i can see, e.g. market risk and credit risk result.

R/imports.R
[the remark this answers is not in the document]
maurolepore: As a maintainer I believe you're welcome to define your preference and I'm happy to follow. I take this as an opportunity to explain my own preferences, which you'll see baked into most r2dii packages on CRAN.

1. The namespace of the package is insulated from the global environment. If the user defines a function with the same name (e.g. `abort <- function() stop("Oh no!")`) the functions in this package will ignore it and still find the one that comes from `#' @importFrom pkg fun`.

2. Great you're thinking on a package-wide convention. Once you define those conventions, I'm more than happy to follow. I don't usually need to define any of my preferences other than saying something like "I use style, patterns, and conventions from the tidyverse team".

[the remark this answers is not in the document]
MirijaHa: @maurolepore  I find aspect 1. very convincing, so let us follow this convention
@jacobvjk There is basically 3 ways to make functions available

1. using :: so dplyr::mutate
2. defining all imports in imports (mauros suggestion)
3. defining imports not in imports.R but in function definitions

I would be against 3, since it calls for clutter and inconsistency, but am fine with 2. :)

R/st_bonds.R
jacobvjk: ah okay, if that is the case, that would indeed be hard to digest

R/st_bonds.R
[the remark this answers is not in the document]
maurolepore: For now I think it's easiest to be consistent with `run_stress_test()` and remove the `data` argument.

For later, I think CRAN forbids writing anywhere other than the session's temporary directory -- unless the user explicitely gives the path. I need to check in CRAN policy's but if that's right then we must not write data to a path hidden behind an envvar. In short, we may get away with reading data from a path in an envvar but not with writing data to it.

--

RE ease of use, I'm pretty sure I had a similar discussion for every package, and I'm sure other developers had it too. I'll find a good way to explain this and maybe write a blog post. 

If you're dealing with new users, then I think it's best to steer them in the right direction right away. Teaching them how to setup an environmental variable so they can read and write data doesn't do users a big favor. They won't be able to reuse that knowledge in the most widely used packages, e.g. the tidyverse. Soon enough R users have no choice other than learning how to read and write data from their system, and to manipulate spreadsheet-like data structure (i.e. dataframes). 

If your users don't want to do that, then they should not interact with the R package directly, but instead with a wrapper like a shiny app. The package remains complient with good design patterns, while the app gives users the interface they feel more comfirtable with.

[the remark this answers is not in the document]
MirijaHa: - remove data: alright
- CRAN forbids writing: True, so that means we  do not even have to discuss env vars further but will get rid of them very early on in CRAN ready version
- RE ease of use: I am not entirely sure I agree with the line of arguments. Users who use our code do not chose to use R and are also not intending to use if further or learn it. They only use it since we do not give them a more suitable option. So while we do not  provide them with a wrapper but force them to use the code directly I think we should put convenience for beginners very high on our priority list.

DESCRIPTION
```diff
@@ -35,6 +35,7 @@ Imports:
     rlang,
     stringr,
     tibble,
+    vctrs,
```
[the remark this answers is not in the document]
MirijaHa: true that, closing comment

maurolepore:
I'm converting to a draft because we agreed to not merge until at least past the DRY run.

maurolepore:
Closing for now, so it's easier to see what's really on the pipeline.

PR #168, opened 2021-11-08
2723 mismatch plan carsten

jacobvjk:
closes ADO 2723: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2723

This PR:
- ensures that the loan data prep calculates exposures with the same reference portfolio size as is later used to scale the company level impact back to the overall loan book.
- previously we erroneously used different denominators (raw loan book for getting loan exposure, matched loan book to rescale company shocks to loan book)
- the loan change check fails expectedly for all results that make use of the exposure. It does not fail for annual and overall pd changes. This is a good sign, since these two results are independent from portfolio exposures which means the calculation of risk on the company level remains unaffected.
- The sector exposures in the results of the new calculation match the sector level exposures in the matched loan book. Previously the result exposures were significantly smaller.


MirijaHa (approved):
LGTM

PR #169, opened 2021-11-08
Report mismatches

MirijaHa:
This closes only parts of https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2180 as agreed.
Change checks pass.
For reference, we lost rows fo the following number of companies (in theory we might lose only some and not all values rows per company when merging on price data).
bonds: 6 out of 193 companies 
equity: 44 out of 319
loans: 9 out of 328


jacobvjk (approved):
lgtm

PR #170, opened 2021-11-08
Fix 'false' must be a double vector

### LINKED ISSUE #167: run_stress_test() errors if term is integer
`term` is likely too picky. It works with the double 1 but fails with the integer 1L.

``` r
devtools::load_all()

# Works with doubles
x <- capture.output(run_stress_test("bonds", term = 1))
#> Warning: Removed 21 rows because negative production compensation targets were set in late and sudden.
#>                    Negative absolute production is impossible

# Fails with integers
x <- capture.output(run_stress_test("bonds", term = 1L))
#> Warning: Removed 21 rows because negative production compensation targets were set in late and sudden.
#>                    Negative absolute production is impossible
#> Error: Problem with `mutate()` column `term`.
#> ℹ `term = dplyr::if_else(.data$term > 5, 5, .data$term)`.
#> x `false` must be a double vector, not an integer vector.
```

<sup>Created on 2021-11-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore:
- Closes #167

This PR fixes the bug in place with the smallest diff I could come up with. 

It does not rule out similar bugs with other arguments. For that we may convert all numeric arguments to an expected type (say integer or double), then we'll know that type will be stable everywhere. That could be in a follow up PR but first I'll investigate if indeed other arguments suffer the same problem.

--

Update:

I see no bug with any of the arguments which default looks like an integer -- `terminal_value`, `div_netprofit_prop_coef`, and `shock_year`. So I won't follow up. 

This is ready for review.

### DIFF
--- a/R/company_expected_loss.R
+++ b/R/company_expected_loss.R
@@ -62,7 +62,7 @@ company_expected_loss <- function(data,
     dplyr::mutate(
       exposure_at_default = .data$asset_portfolio_value * .data$percent_exposure
     ) %>%
-    dplyr::mutate(term = dplyr::if_else(.data$term > 5, 5, .data$term))
+    dplyr::mutate(term = dplyr::if_else(.data$term > 5, 5, as.double(.data$term)))
 
   data <- data %>%
     dplyr::mutate(lgd = loss_given_default)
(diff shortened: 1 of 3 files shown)


MirijaHa (approved):
thx a lot

PR #172, opened 2021-11-08
Avoid NOTE: no visible binding for '<<-'

maurolepore:
This PR avoids the R CMD check NOTE: 

```
no visible binding for '<<-'﻿
```

The files it adds to .Rbuildignore are not used in the
package code. But they seem necessary for legacy code 
that uses the source code directly from the repo that
hosts this package.

REVIEWER: Please confirm this does not break code that
depends on these files.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,3 +1,9 @@
+# FIXME: Avoid NOTE: no visible binding for '<<-' assignment
+# These files are not used by the package but they are by legacy code that
+# uses the repository that hosts the package.
+^R/set_params_st\.R$
+^tests/testthat/test-set_params_st\.R$
+
 ^r2dii\.climate\.stress\.test\.Rproj$
 ^\.Rproj\.user$
 ^README\.Rmd$


MirijaHa:
> This PR avoids the R CMD check NOTE:
> 
> ```
> no visible binding for '<<-'﻿
> ```
> 
> The files it adds to .Rbuildignore are not used in the package code. But they seem necessary for legacy code that uses the source code directly from the repo that hosts this package.
> 
> REVIEWER: Please confirm this does not break code that depends on these files.

From what I see the webtool script gets the functions via source and devtools::load all https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/21831ec00184615c789e94f8055fe7830cf3fb10/web_tool_stress_test.R#L53, so excluding the file from build should not be a problem. However I myself know nothing about how the webtool scripts are used, so will tag @jacobvjk  for review

jacobvjk (approved):
lgtm
still works in the web tool

PR #174, opened 2021-11-09
2559 missing var company plan carsten

jacobvjk:
closes ADO 2559 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2559

This PR:
- removes exposures that have zero value in the portfolio
- this means that we will not try to aggregate company/tech level results to the portfolio and will no longer generate company level NAs based on this
- change checks pass, except for the 0 exposures that are removed
- portfolio level and non-exposure related results stay equal, which is expected

### DIFF
--- a/R/calculate.R
+++ b/R/calculate.R
@@ -121,5 +121,18 @@ calculate_exposure_by_technology_and_company <- function(asset_type,
      composite_unique_cols = names(.)
    )
 
+  all_exposures <- length(unique(exposure_by_technology_and_company$company_name))
+
+  wrangled_pacta_results <- exposure_by_technology_and_company %>%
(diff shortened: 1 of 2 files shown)


MirijaHa (commented):
- Can  this only be checked here and not as part of report_company_drops?
- if not maybe it would be nice to adjust wording and formatting to the approach taken in report_dropped_company_names still?

jacobvjk:
> * Can  this only be checked here and not as part of report_company_drops?
> * if not maybe it would be nice to adjust wording and formatting to the approach taken in report_dropped_company_names still?

I am filtering here, so I think using report_dropped_company_names is a bit complicated, but I will adapt the message to match the style

MirijaHa (approved):

PR #176, opened 2021-11-09
fix exclusion

MirijaHa:
this closes https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2831

Change checks run through.
However I made a, to my knowledge, weird observation @jacobvjk . Namely when comparing results for equity for run with company_exclusion = TRUE vs company_exclusion ==FALSE I got results with the same number of rows but different values (all result types). Is that expected? I would have assumed that company ecxlusion would either also number of rows for equity_results_company if it is an effect on results.

### DIFF
--- a/R/check_and_filter_data.R
+++ b/R/check_and_filter_data.R
@@ -26,8 +26,12 @@ check_and_filter_data <- function(st_data_list, start_year, end_year,
     dplyr::filter(.data$technology %in% .env$technologies_lookup) %>%
     dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$end_year))
 
-  excluded_companies_filtered <- st_data_list$excluded_companies %>%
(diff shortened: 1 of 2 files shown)


jacobvjk:
> this closes https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2831
> 
> Change checks run through. However I made a, to my knowledge, weird observation @jacobvjk . Namely when comparing results for equity for run with company_exclusion = TRUE vs company_exclusion ==FALSE I got results with the same number of rows but different values (all result types). Is that expected? I would have assumed that company ecxlusion would either also number of rows for equity_results_company if it is an effect on results.

this is expected. you will see a column `exclusion` that is a boolean and indicates whether or not the company/tech combination was considered in the calculations or not. but rows are not removed.
We may think about changing this, I am open to that.

jacobvjk (approved):
lgtm

PR #179, opened 2021-11-09
Remove suffix _arg from argument names, not output

maurolepore:
This remove the suffix '_arg' from the signature of `run_stress_test()`
but leave it in the name of the output columns.

I aimed to minimize the diff. 

I think we should refactor as I think there is a simpler way to the job. 
In general, I think the output is easier to grow rowwise than colwise.
If we need columns we can pivot_wider() at the end. But I suspect the 
longer format is easier to pass to ggplot, and filter with
dplyr::filter().

### DIFF
--- a/R/run_stress_test.R
+++ b/R/run_stress_test.R
@@ -5,44 +5,44 @@
 #' vector of values to one (and only one) of the detail arguments. This will
 #' result in running the analysis multiple times in a row with the argument
 #' varied.
-#' NOTE: argument `asset_type_arg` is not iterateable.
+#' NOTE: argument `asset_type` is not iterateable.
 #'
-#' @param asset_type_arg String holding asset_type, for allowed value compare
+#' @param asset_type String holding asset_type, for allowed value compare
 #'   `asset_types_lookup`.
-#' @param lgd_senior_claims_arg Numeric, holding the loss given default for senior
+#' @param lgd_senior_claims Numeric, holding the loss given default for senior
 #'   claims, for accepted value range check `lgd_senior_claims_range_lookup`.
-#' @param lgd_subordinated_claims_arg Numeric, holding the loss given default for
+#' @param lgd_subordinated_claims Numeric, holding the loss given default for
 #'   subordinated claims, for accepted value range check
 #'   `lgd_subordinated_claims_range_lookup`.
-#' @param terminal_value_arg Numeric. A ratio to determine the share of the
+#' @param terminal_value Numeric. A ratio to determine the share of the
 #'   discounted value used in the terminal value calculation beyond the
 #'   projected time frame. For accepted range compare `terminal_value_range_lookup`.
-#' @param risk_free_rate_arg Numeric that indicates the risk free rate of interest.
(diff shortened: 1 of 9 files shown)


R/write_results.R
```diff
@@ -26,6 +26,7 @@ write_stress_test_results <- function(results, expected_loss,
 
   results_path <- file.path(get_st_data_path("ST_PROJECT_FOLDER"), "outputs")
 
+  sensitivity_analysis_vars <- paste0(sensitivity_analysis_vars, "_arg")
```
maurolepore, line 29: We now have to tell the validation function to expect the bare name of the argument plus the suffix "_arg".

R/run_stress_test.R
```diff
@@ -86,6 +86,7 @@ run_stress_test_iteration <- function(n, args_tibble) {
   arg_list_row <- arg_tibble_row %>%
     as.list()
 
+  arg_tibble_row <- dplyr::rename_with(arg_tibble_row, ~paste0(.x, "_arg"))
```
maurolepore, line 89: Here is the trick.

nonstandard/QA_scripts/change_check_cb.R
```diff
@@ -81,11 +81,11 @@ check
...
-  bonds_results_port <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_port_shock_year_arg.csv"))
-  bonds_expected_loss <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_comp_el_shock_year_arg.csv"))
-  bonds_annual_pd_changes_sector <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_sector_pd_changes_annual_shock_year_arg.csv"))
-  bonds_overall_pd_changes_sector <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_sector_pd_changes_overall_shock_year_arg.csv"))
+  bonds_results_company <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_comp_shock_year.csv"))
+  bonds_results_port <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_port_shock_year.csv"))
+  bonds_expected_loss <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_comp_el_shock_year.csv"))
+  bonds_annual_pd_changes_sector <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_sector_pd_changes_annual_shock_year.csv"))
+  bonds_overall_pd_changes_sector <- readr::read_csv(file.path(results_path, "stress_test_results_bonds_sector_pd_changes_overall_shock_year.csv"))
```
maurolepore, line 88: @MirijaHa, please review this carefully. I did this changes automatically and I'm not sure if they are okay. The same applies to all other changes in nonstandard/.

MirijaHa (approved):
Reviewed carefully and ran pre post checks for bonds for iteration and non-iteration case. LGTM, thx again 🙏 , I will be so free to merge it myself to unblock the following stories

PR #180, opened 2021-11-10
move r2dii.plot to suggst

MirijaHa:
microchange to force users to install less deps.
config and ggplot are also not needed for user workflows, but since the are used in webtool they cannot be removed from imports yet.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,6 @@ Imports:
     purrr,
     r2dii.analysis,
     r2dii.data,
-    r2dii.plot,
     readr,
     readxl,
     rlang,
@@ -39,6 +38,7 @@ Imports:
     withr,
     zoo
 Suggests: 
+    r2dii.plot,
     testthat (>= 3.0.0)
 Config/testthat/edition: 3
 Encoding: UTF-8


jacobvjk (approved):
lgtm

maurolepore:
For the packages that the webtool needs:

Maybe we can do something like this:

1. Move those packages to "Suggests:".

2. Confirm it's the webtool, by checking if "package:r2dii.climate.stress.test" is not in `search()` (so the user did not run `library(r2dii.climate.stress.test)`.

3. Use `rlang::check_installed()` with the packages required only for the web tool.

MirijaHa:
> For the packages that the webtool needs:
> 
> Maybe we can do something like this:
> 
> 1. Move those packages to "Suggests:".
> 2. Confirm it's the webtool, by checking if "package:r2dii.climate.stress.test" is not in `search()` (so the user did not run `library(r2dii.climate.stress.test)`.
> 3. Use `rlang::check_installed()` with the packages required only for the web tool.

Unfortunately I know no details on the webtool interaction between PACTA code and stresstesting so I cannot say if this would be safe enough. If I understand correctly that would mean that it respective packages were not around the runs would fail in the rlang::check_installed step, I do not know if we can risk that.

PR #182, opened 2021-11-10
No more env vars

MirijaHa:
This closes ADO https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2556 , exposes data path args to users.
For internal convenience I use get_st_data_path in function calls for tests ands change checks.
In drafr mode till I have tested thoroughtly.

### DIFF
--- a/R/read_and_prepare.R
+++ b/R/read_and_prepare.R
@@ -10,25 +10,25 @@
 #' @param scenario_geography_filter String holding geographic asset locations to
 #'   be considered in analysis.
 #' @param asset_type Type of asset, can be bonds, loans or equity.
+#' @param path String holding path to data.
 #'
 #' @return A list holding prepared project agnostic data.
 read_and_prepare_project_agnostic_data <- function(start_year, end_year, company_exclusion,
-                                                   scenario_geography_filter, asset_type) {
+                                                   scenario_geography_filter, asset_type,
+                                                   path) {
   if (!asset_type %in% asset_types_lookup) {
     stop("Invalid asset type provided.")
   }
 
-  data_location <- get_st_data_path()
-
   capacity_factors_power <- read_capacity_factors(
-    path = file.path(data_location, "capacity_factors_WEO_2020.csv"),
+    path = file.path(path, "capacity_factors_WEO_2020.csv"),
     version = "new"
   )
 
   if (company_exclusion) {
-    validate_file_exists(file.path(data_location, "exclude-companies.csv"))
+    validate_file_exists(file.path(path, "exclude-companies.csv"))
     excluded_companies <- readr::read_csv(
-      file.path(data_location, "exclude-companies.csv"),
+      file.path(path, "exclude-companies.csv"),
       col_types = readr::cols(
         company_name = "c",
         technology = "c"
@@ -39,7 +39,7 @@ read_and_prepare_project_agnostic_data <- function(start_year, end_year, company
   }
 
   df_price <- read_price_data(
-    path = file.path(data_location, paste0("prices_data_", price_data_version_lookup, ".csv")),
+    path = file.path(path, paste0("prices_data_", price_data_version_lookup, ".csv")),
     version = "old",
     expected_technologies = technologies_lookup
   ) %>%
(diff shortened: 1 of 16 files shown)


MirijaHa:
Change checks pass.
The changes make it a bit more fiddly, I wonder if we should first refactor run_stress_test further before adding this. Then again it works, and it would be highly wishable that this is on master on monday for the 2nd dryrun.

R/read_and_prepare.R
```diff
@@ -10,25 +10,25 @@
 #' @param scenario_geography_filter String holding geographic asset locations to
 #'   be considered in analysis.
 #' @param asset_type Type of asset, can be bonds, loans or equity.
+#' @param path String holding path to data.
 #'
 #' @return A list holding prepared project agnostic data.
 read_and_prepare_project_agnostic_data <- function(start_year, end_year, company_exclusion,
-                                                   scenario_geography_filter, asset_type) {
+                                                   scenario_geography_filter, asset_type,
+                                                   path) {
```
maurolepore, line 18: The argument that provides the data is usually in position 1. This is particularly the case for functions that read data

``` r
readers <- c(read.csv, readr::read_csv)
lapply(readers, \(x) formals(x)[1])
#> [[1]]
#> [[1]]$file
#> 
#> 
#> 
#> [[2]]
#> [[2]]$file
```

<sup>Created on 2021-11-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/read_and_prepare.R
```diff
@@ -79,16 +79,16 @@ read_and_prepare_project_agnostic_data <- function(start_year, end_year, company
...
 read_and_prepare_project_specific_data <- function(asset_type, calculation_level,
                                                    time_horizon, scenario_geography_filter,
                                                    scenarios_filter, equity_market_filter,
-                                                   term) {
-  path <- file.path(get_st_data_path("ST_PROJECT_FOLDER"), "inputs", paste0(stringr::str_to_title(asset_type), "_results_", calculation_level, ".rda"))
+                                                   term, path) {
 
   pacta_results <- read_pacta_results(
-    path = path,
+    path = file.path(path, "inputs", paste0(stringr::str_to_title(asset_type), "_results_", calculation_level, ".rda")),
```
maurolepore, line 91: An alternative:

``` r
stringr::str_to_title(c("bonds", "loans"))
#> [1] "Bonds" "Loans"
tools::toTitleCase(c("bonds", "loans"))
#> [1] "Bonds" "Loans"
```

<sup>Created on 2021-11-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/run_stress_test.R
```diff
@@ -34,6 +38,8 @@
 #' @return NULL
 #' @export
 run_stress_test <- function(asset_type,
+                            data_path_project_specific,
+                            data_path_project_agnostic,
```
maurolepore, line 42: I think it's rare that a function would have two arguments for input. Commonly you have only one. Maybe take a single path to the single parent directory containing two subdirectoreis -- one for project specific and the other one for project agnostic data?

maurolepore: Also, the biggest concern is the output directory. Right now I think it's unclear to the user how to set outputs independently from inputs. That will make the function a lot easier to test, since we'll be able to send outputs to a self-destroyable directory like `tempdir()`.

tests/testthat/test-run_stress_test.R
```diff
@@ -6,9 +6,13 @@ test_that("output includes argument names with suffix '_arg'", {
...
-  x <- suppressWarnings(capture.output(
-    run_stress_test("bonds", term = 1:2)
-  ))
+x <- suppressWarnings(capture.output(
+  run_stress_test("bonds",
+    data_path_project_agnostic = get_st_data_path(),
+    data_path_project_specific = get_st_data_path("ST_PROJECT_FOLDER"),
+    term = 1:2
+  )
+))
```
maurolepore, line 15: :heart:

maurolepore (approved):
LGTM. Yay!!! Goodbye envvars!

I leave some non-blocking comments.

R/run_stress_test.R
MirijaHa: I would rather not combine  project agnostic and project independent files under a common subdirectory. This is for practical reasons:

MirijaHa: this way however people would have to pass also the name of the project to the function though, at least if we want to avoid that ST_INPUTS master has to be copy pasted into every single projects folder

PR #190, opened 2021-11-11
No env prep

MirijaHa:
embarrassingly  I forgot to adjust this for the loan book prep initially, here it is


jacobvjk (approved):
even more embarrassingly, I found that and decided to only add one of the two to the vignette.. damn.
lgtm

jacobvjk (approved):
dazzling!
LGTM

PR #189, opened 2021-11-10
New st_arguments: Search defaults and valid values

maurolepore:
Closes #188

This PR adds a new dataset to help users find the default and valid inputs to the arguments of `run_stress_test()`.

I did not remove other ways to achieve the same (i.e. using the individual `<argument name>_range_lookup` objects -- to avoid breaking changes and to propose this idea with minimum effort. If this is accepted, then I can work on adapting the existing code to this new dataset.

``` r
r2dii.climate.stress.test::st_arguments
#> # A tibble: 10 × 3
#>    name                    default range               
#>    <chr>                   <chr>   <chr>               
#>  1 asset_type              <NA>    equity, bonds, loans
#>  2 lgd_senior_claims       0.45    0.3, 0.6            
#>  3 lgd_subordinated_claims 0.75    0.6, 0.9            
#>  4 terminal_value          0       0, 0.1              
#>  5 risk_free_rate          0.02    0, 0.05             
#>  6 discount_rate           0.02    -0.01, 0.05         
#>  7 div_netprofit_prop_coef 1       0.8, 1              
#>  8 shock_year              2030    2025, 2040          
#>  9 term                    2       1, 10               
#> 10 company_exclusion       TRUE    TRUE, FALSE
```

<sup>Created on 2021-11-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>﻿

PR #192, opened 2021-11-11
tidy errors

MirijaHa:
As we noted in dryrun that current error messages are not particularly helpful I am adjusting the most likely errors that will require user action to the admittedly more informative tidyverse convention.
WIll add a follow up PR for input argument validations when @maurolepore s suggestion here  https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/189 is implemented, as this will simplify adapting the errors for input args validation.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -66,14 +66,22 @@ path_dropbox_2dii <- function(...) {
 #' Before performing an operation on a file assumed to be found in a given
 #' directory, validate this file exists and give indicative error if not.
 #'
-#' @param path Character vector indicating the directory of a file
+#' @param path Character vector indicating the directory of a file.
 #'
-#' @return A boolean.
+#' @return NULL
 #'
 #' @export
 validate_file_exists <- function(path) {
   valid_file_path <- file.exists(file.path(path))
-  stopifnot(valid_file_path)
+
+  if (!valid_file_path) {
+    rlang::abort(c(
+      "Detected invalid file path.",
+      x = glue::glue("Invalid file path: {file.path(path)}."),
+      i = "Did you set path to data correctly?."
+    ))
+  }
+  invisible()
 }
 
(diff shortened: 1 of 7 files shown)


jacobvjk (approved):
lgtm

R/utils.R
```diff
@@ -66,14 +66,22 @@ path_dropbox_2dii <- function(...) {
...
-  stopifnot(valid_file_path)
+
+  if (!valid_file_path) {
+    rlang::abort(c(
+      "Detected invalid file path.",
+      x = glue::glue("Invalid file path: {file.path(path)}."),
+      i = "Did you set path to data correctly?."
+    ))
+  }
+  invisible()
```
maurolepore, line 84: This should be path.

> If a function is called primarily for its side-effects, it should invisibly return a useful output. If there’s no obvious output, return the first argument. This makes it possible to use the function with in a pipeline.

-- https://design.tidyverse.org/out-invisible.html


```r
"/a/path" %>%
  validate_file_exists() %>%
  do_something_else_with_the_path()
```

R/utils.R
```diff
@@ -66,14 +66,22 @@ path_dropbox_2dii <- function(...) {
...
+#' @return NULL
 #'
 #' @export
 validate_file_exists <- function(path) {
   valid_file_path <- file.exists(file.path(path))
-  stopifnot(valid_file_path)
+
+  if (!valid_file_path) {
+    rlang::abort(c(
+      "Detected invalid file path.",
```
maurolepore, line 79: Nit:

> If the cause of the problem is clear, use “must”:
-- https://style.tidyverse.org/error-messages.html#problem-statement

Suggested:

"`path` must point to an existing file."

maurolepore (commented):
Thanks! I make small nit suggestions.

PR #194, opened 2021-11-11
Update website's reference

maurolepore:
Show user-facing functions only.

We can tweak the website a bit more and make it similar to other packages of the r2dii family, e.g. https://2degreesinvesting.github.io/r2dii.data/

image

--

And we can set it up so that we host the released version at the normal URL, and the development version at ULR/dev:

image


MirijaHa (approved):
thx

PR #195, opened 2021-11-11
2393 check distincts

jacobvjk:
closes ADO 2393: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2393

I did not touch the ones exclusively related to the web tool or litigation risk
change checks passed

### DIFF
--- a/R/extend_scenario_trajectory.R
+++ b/R/extend_scenario_trajectory.R
@@ -78,7 +78,7 @@ extend_scenario_trajectory <- function(data,
       composite_unique_cols = c(
         "year", "investor_name", "portfolio_name", "id", "company_name",
         "ald_sector", "technology", "scenario", "allocation",
-        "scenario_geography", "source", "units", "direction"
+        "scenario_geography", "source", "units"
       )
     )
 
--- a/R/extend_scenario_trajectory.R
+++ b/R/extend_scenario_trajectory.R
@@ -55,7 +55,6 @@ extend_scenario_trajectory <- function(data,
       .data$allocation, .data$year, .data$scenario, .data$plan_tech_prod,
       .data$scen_tech_prod
     ) %>%
-    # TODO: explain magic number
     dplyr::filter(.data$year <= .env$start_analysis + .env$time_frame) %>%
     tidyr::complete(
       year = seq(.env$start_analysis, .env$end_analysis),
@@ -70,9 +69,17 @@ extend_scenario_trajectory <- function(data,
     )
 
   data <- data %>%
+    # ADO 2393 - The join cols should be extended to cover the source
     dplyr::inner_join(
       scenario_data,
       by = c("ald_sector", "technology", "scenario_geography", "scenario", "year")
+    ) %>%
+    report_all_duplicate_kinds(
+      composite_unique_cols = c(
+        "year", "investor_name", "portfolio_name", "id", "company_name",
+        "ald_sector", "technology", "scenario", "allocation",
+        "scenario_geography", "source", "units", "direction"
+      )
     )
 
   data <- data %>%
--- a/R/extend_scenario_trajectory.R
+++ b/R/extend_scenario_trajectory.R
@@ -73,8 +73,7 @@ extend_scenario_trajectory <- function(data,
     dplyr::inner_join(
       scenario_data,
       by = c("ald_sector", "technology", "scenario_geography", "scenario", "year")
-    ) %>%
-    dplyr::distinct_all()
+    )
 
   data <- data %>%
     dplyr::group_by(
--- a/R/read_capacity_factors.R
+++ b/R/read_capacity_factors.R
@@ -63,12 +63,13 @@ read_capacity_factors <- function(path = NULL,
     )
 
   } else {
+    # ADO 2393 - this should keep the source and filter based on global settings
+    # not in this hard coded manner
     data <- data %>%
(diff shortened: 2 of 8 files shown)


R/extend_scenario_trajectory.R
```diff
@@ -73,8 +73,7 @@ extend_scenario_trajectory <- function(data,
     dplyr::inner_join(
       scenario_data,
       by = c("ald_sector", "technology", "scenario_geography", "scenario", "year")
-    ) %>%
-    dplyr::distinct_all()
+    )
```
MirijaHa, line 76: just to go sure, there is no possible scenario where the distinct_all here is needed? When in doubt I would recommen using report_all_duplicate kinds on it. The fucntion has been changed to invisibly return the input, so you could simply pipe it in with %>% reprt_all_duplicate_kinds(names(.))

R/run_prep_calculation_loans.R
```diff
@@ -287,9 +280,7 @@ run_prep_calculation_loans <- function(data_path_project_specific,
     dplyr::rename(
       production_unweighted = .data$production
     )  %>%
-    dplyr::mutate(technology_share = round(.data$technology_share, 8)) %>% # rounding errors can lead to duplicates
-    # TODO: why distinct_all?
-    dplyr::distinct_all()
+    dplyr::mutate(technology_share = round(.data$technology_share, 8))# rounding errors can lead to duplicates
```
MirijaHa, line 283: for my understanding, are the duplicates only on technology share? That should be unproblematic, in that case I woudl rather remove the comment, sicne it has a bit of a fix me character to it.

tests/testthat/test-calculate_annual_pd_changes.R
```diff
@@ -29,13 +29,17 @@ test_that("PD_changes point in expected direction", {
   test_end_of_analysis <- 2040
   test_risk_free_rate <- 0.05
 
+  # Expected direction of PD change depends on direction of expected value change
+  # This can be extracted from NPV columns of the annual profit data, which is
+  # equal for all years
+  filter_year <- min(test_data$year, na.rm = TRUE)
   expected_direction <- test_data %>%
-    dplyr::distinct(
+    dplyr::filter(.data$year == filter_year) %>%
```
MirijaHa, line 37: nice

R/read_capacity_factors.R
```diff
@@ -62,12 +63,13 @@ read_capacity_factors <- function(path = NULL,
     )
 
   } else {
+    # ADO 2393 - this should keep the source and filter based on global settings
+    # not in this hard coded manner
     data <- data %>%
       dplyr::select(
         .data$scenario, .data$scenario_geography, .data$technology,
         .data$year, .data$capacity_factor
       ) %>%
```
MirijaHa, line 72: maybe pipe in report_all_duplicate_kinds if in any thinkable way needed

R/run_prep_calculation_loans.R
```diff
@@ -185,7 +176,8 @@ run_prep_calculation_loans <- function(data_path_project_specific,
       .data$comp_loan_size_credit_limit,
       .data$loan_size_credit_limit_currency
     ) %>%
-    # TODO why distinct_all?
+    # ADO 2393 - somehow the summarise call does not reduce the number of rows.
```
MirijaHa, line 179: that is a bit of a scary comment. If your expectation was that summarise should reduced rows you might want to sharply look at it once more, maybe there are some unexpected patterns in the data set

MirijaHa (commented):
I am very happy about every distinct_all that disappeared, left some small comments. To go sure, did you verify that change checks pass and that the prepared loanbook remains unaltered?

R/run_prep_calculation_loans.R
jacobvjk: I have to admit I do not fully remember where that comment came from and my guess is that it may be a wrong comment altogether. So maybe I will just remove it and also report duplicated on CUC cols

R/read_capacity_factors.R
jacobvjk: will add here. in these functions around the extending and wrangling the scenarios, I also think it would be good to look at passing the scenario source soon. We get that for pacta now and we would also be able to get that from all scenario related files. This would make it more robust.

R/extend_scenario_trajectory.R
jacobvjk: done

MirijaHa: to me it looked like you havent. Or just not in this fucniton because uneccesray?

R/run_prep_calculation_loans.R
jacobvjk: found the issue, weird aggregation.. came to the same result, but now it is much more concise

jacobvjk:
> I am very happy about every distinct_all that disappeared, left some small comments. To go sure, did you verify that change checks pass and that the prepared loanbook remains unaltered?

made the changes as pointed out, result changes still pass

R/extend_scenario_trajectory.R
MirijaHa: imo direction is not part of CUC

jacobvjk: you are right!

MirijaHa (approved):

PR #197, opened 2021-11-12
Output path

MirijaHa:
This PR go uncomfortably big, I am slightly sorry:

- included a sneaky refactoring to pipe in validate_file_exisits
- added output_path (run_stress_test) and data_prep_output_path (run_prep_loan_book)
- change checks passed
- new loan book input data were generated successfully

Some remarks:

- i wonder if to combine all 3 path in a list arg data_paths to make the interface leaner. However that would kind of make the fact that we write to an output dir less obvious, so not sure
- the output path for data_prep is the input path in the current approach. This might be confusing. I would rather not adapt our file approach here though, because that would be too much overhead to adapt all docs, after all in the hope that we can get rid of run_prep_loan_book at some point. Made it somewhat clear in the docs though
- it is becoming increasingly painful to change the interface due to the needed adjustments in readme, vignette, tests etc, we should keep an eye on that.

### DIFF
--- a/R/read_and_prepare.R
+++ b/R/read_and_prepare.R
@@ -86,7 +86,7 @@ read_and_prepare_project_specific_data <- function(asset_type, calculation_level
                                                    scenarios_filter, equity_market_filter,
                                                    term, path) {
   pacta_results <- read_pacta_results(
-    path = file.path(path, "inputs", paste0(stringr::str_to_title(asset_type), "_results_", calculation_level, ".rda")),
+    path = file.path(path, paste0(stringr::str_to_title(asset_type), "_results_", calculation_level, ".rda")),
     level = calculation_level
   )
 
@@ -104,7 +104,7 @@ read_and_prepare_project_specific_data <- function(asset_type, calculation_level
     # TODO: next version to allow term input on holding/company level
     dplyr::mutate(term = term)
 
-  sector_exposures <- read_sector_exposures(file.path(path, "inputs", "overview_portfolio.rda")) %>%
+  sector_exposures <- read_sector_exposures(file.path(path, "overview_portfolio.rda")) %>%
     wrangle_and_check_sector_exposures(asset_type = asset_type)
   # TODO: potentially convert currencies to USD or at least common currency
 
--- a/R/read_and_prepare.R
+++ b/R/read_and_prepare.R
@@ -26,14 +26,13 @@ read_and_prepare_project_agnostic_data <- function(start_year, end_year, company
   )
 
   if (company_exclusion) {
-    validate_file_exists(file.path(path, "exclude-companies.csv"))
-    excluded_companies <- readr::read_csv(
-      file.path(path, "exclude-companies.csv"),
-      col_types = readr::cols(
-        company_name = "c",
-        technology = "c"
+    excluded_companies <- validate_file_exists(file.path(path, "exclude-companies.csv")) %>%
+      readr::read_csv(
+        col_types = readr::cols(
+          company_name = "c",
+          technology = "c"
+        )
       )
-    )
   } else {
     excluded_companies <- NULL
   }
@@ -59,7 +58,7 @@ read_and_prepare_project_agnostic_data <- function(start_year, end_year, company
   financial_data <- read_financial_data(
     path = file.path(get_st_data_path(), "prewrangled_financial_data_stress_test.csv")
   ) %>%
-  check_financial_data(asset_type = asset_type)
+    check_financial_data(asset_type = asset_type)
 
   return(list(
     capacity_factors_power = capacity_factors_power,
@@ -86,7 +85,6 @@ read_and_prepare_project_specific_data <- function(asset_type, calculation_level
                                                    time_horizon, scenario_geography_filter,
                                                    scenarios_filter, equity_market_filter,
                                                    term, path) {
-
   pacta_results <- read_pacta_results(
     path = file.path(path, "inputs", paste0(stringr::str_to_title(asset_type), "_results_", calculation_level, ".rda")),
     level = calculation_level
@@ -130,9 +128,8 @@ read_and_prepare_project_specific_data <- function(asset_type, calculation_level
 #'
 #' @return A tibble holding sector exposure data
 read_sector_exposures <- function(path) {
-  validate_file_exists(path)
-
-  sector_exposures <- readr::read_rds(path)
+  sector_exposures <- validate_file_exists(path) %>%
+    readr::read_rds()
 
   validate_data_has_expected_cols(
     data = sector_exposures,
--- a/R/read_scenario_data.R
+++ b/R/read_scenario_data.R
@@ -7,19 +7,20 @@
 #' @return A tibble holding scenario data.
 read_scenario_data <- function(path) {
 
-  validate_file_exists(path)
-
-  scenario_data <- readr::read_csv(path, col_types = readr::cols(
-    scenario_source = "c",
-    scenario_geography = "c",
-    scenario = "c",
-    ald_sector = "c",
-    units = "c",
-    technology = "c",
-    year = "d",
-    direction = "c",
-    fair_share_perc = "d"
-  ))
+  scenario_data <- validate_file_exists(path) %>%
+    readr::read_csv(
+      col_types = readr::cols(
+        scenario_source = "c",
+        scenario_geography = "c",
+        scenario = "c",
+        ald_sector = "c",
+        units = "c",
+        technology = "c",
+        year = "d",
+        direction = "c",
+        fair_share_perc = "d"
+      )
+    )
 
   validate_data_has_expected_cols(
     data = scenario_data,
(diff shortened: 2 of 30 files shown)


R/read_and_prepare.R
```diff
@@ -26,14 +26,13 @@ read_and_prepare_project_agnostic_data <- function(start_year, end_year, company
   )
 
   if (company_exclusion) {
-    validate_file_exists(file.path(path, "exclude-companies.csv"))
-    excluded_companies <- readr::read_csv(
-      file.path(path, "exclude-companies.csv"),
-      col_types = readr::cols(
-        company_name = "c",
-        technology = "c"
+    excluded_companies <- validate_file_exists(file.path(path, "exclude-companies.csv")) %>%
```
jacobvjk, line 29: coolio

R/run_prep_calculation_loans.R
```diff
@@ -1,13 +1,20 @@
 #' Run stress testing data preparation for loans
 #'
+#'
 #' @inheritParams run_stress_test
+#' @param data_prep_output_path Path where results are written.
+#'   NOTE: This is a workflow that is needed exclusively in preparation of
+#'   [run_stress_test()] for loan books. It creates input data for the stress
+#'   test. A recommended setting is to set `data_prep_output_path` to to the same
```
jacobvjk, line 8: Is it just recommended? If I were to set the output of the prep somewhere else, I could still get the main calc to run, by changing the specific path?

jacobvjk (approved):
very nice solution, also a big fan of the sneaky refactoring here
LGTM
one none blocking question

R/run_prep_calculation_loans.R
```diff
@@ -1,13 +1,20 @@
 #' Run stress testing data preparation for loans
 #'
+#'
 #' @inheritParams run_stress_test
+#' @param data_prep_output_path Path where results are written.
```
maurolepore, line 5: I think the argument name is too long. You rarely see this. I think the should be evocative enough that it jogs your memory, but the details should be not in the name but in the helpfile. While I get there is some extra clarity in a long name, the code overall becomes less clear when they force you or the user to wrap a line in an odd place -- before that line becomes coherent in itself.


```r
# Clearer: The entire idea fits in a single line
# For details see `?function_name()` -- as usual
function_name(argument1 = "blah1", argument2 = "blah2")

# Less clear
here_is_a_long_function_name(
   with_this_long_argument1= "blah1", 
   and_this_other_argument2 = "blah2"
)
```

This gets much worse when you consider more than on function in a nested call or a pipe.

R/run_prep_calculation_loans.R
```diff
@@ -1,13 +1,20 @@
...
 #' @param credit_type Type of credit. For accepted values please compare
 #'   `credit_type_lookup`.
 #'
 #' @return NULL
 #' @export
-run_prep_calculation_loans <- function(data_path_project_specific,
-                                       data_path_project_agnostic,
+run_prep_calculation_loans <- function(input_path_project_specific,
+                                       input_path_project_agnostic,
+                                       data_prep_output_path,
```
maurolepore, line 17: Here is one example of the long-name problem. The wrapping is less clear than this -- I believe:

```r
prep_loans <- function(general_input_path, specific_input_path, output_path) { ...
```

And it would be much clearer if you make each function do one thing and only one thing well:

```r
loans_read <- function(general_path, specific_path) { ...

loans_prep <- function(data) { ...

loans_write <- function(data, path) { ...
```

Note here the `loan_` prefix plays well with autocomplete and makes a clear family versus `st_*()` functions.

R/utils.R
```diff
@@ -68,20 +68,20 @@ path_dropbox_2dii <- function(...) {
 #'
 #' @param path Character vector indicating the directory of a file.
 #'
-#' @return NULL
+#' @return String holding provided `path`.
```
maurolepore, line 71: See how Jenny does it: https://readr.tidyverse.org/reference/write_delim.html#value
In general, it's useful to clarify that the return value is invisible to the user isn't surprised by the lack of output. Here my comment is FYI because I assume this function is only internal.

nonstandard/QA_scripts/qa_cb.R
```diff
@@ -19,8 +19,8 @@ function_paths <- c(
 source_all(function_paths)
 
 inputs_path <- get_st_data_path()
-results_path <- file.path(get_st_data_path("ST_PROJECT_FOLDER"), "outputs")
-graph_path <- file.path(get_st_data_path("ST_PROJECT_FOLDER"), "outputs", "graphs")
+results_path <- get_st_data_path("ST_PROJECT_FOLDER_OUTPUT")
+graph_path <- file.path(get_st_data_path("ST_PROJECT_FOLDER_OUTPUT"), "graphs")
```
maurolepore, line 23: These strings are scattered in multiple places. I think it's best to abstract them in a lil helper to make future changes easier

```r
some_name <- function() "ST_PROJECT_FOLDER_OUTPUT"
```

Then, the day we rename this to say "ST_PROJECT_OUTPUT" the diff will show a single change in the body of `some_function()`. This is the idea that DRY code is easy to maintain.

vignettes/articles/prepare-loans-inputs.Rmd
```diff
@@ -71,8 +74,9 @@ achieve this, instead of the previous command, simply run:
 
 ```{r}
 run_prep_calculation_loans(
-  data_path_project_specific = "/path/to/specific/data",
-  data_path_project_agnostic = "/path/to/agnostic/data",
+  input_path_project_specific = "/path/to/specific/data",
+  input_path_project_agnostic = "/path/to/agnostic/data",
+  data_prep_output_path = "/path/to/specific/data",
   credit_type = "credit_limit"
 )
```
maurolepore, line 81: Run? Prepare? or Calculate? What's the action here?

I suggest `prep_loans()`.

maurolepore (approved):
LGTM! Just come general comments. Thanks!

R/run_prep_calculation_loans.R
MirijaHa: I do not entirely understand the question. I would say it is recommended, because that is easiest, but not enforced, because a user can pick any path, they will jsut have to do some copy paste action then

R/read_scenario_data.R
```diff
@@ -7,19 +7,20 @@
...
+        scenario_source = "c",
+        scenario_geography = "c",
+        scenario = "c",
+        ald_sector = "c",
+        units = "c",
+        technology = "c",
+        year = "d",
+        direction = "c",
+        fair_share_perc = "d"
+      )
```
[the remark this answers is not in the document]
MirijaHa: I had been thinking about doing this in the context of improving our data mgmt strategy, ist is on the wishlist :)

R/run_prep_calculation_loans.R
jacobvjk: okay that's what I meant. Got it, I will strongly suggest it to be used that way in any meeting/document

nonstandard/QA_scripts/qa_cb.R
MirijaHa: on the wishlist

maurolepore (commented):
@MirijaHa, 

I forgot to share my thoughts about some of your remarks:

> i wonder if to combine all 3 path in a list arg data_paths to make the interface leaner.However that would kind of make the fact that we write to an output dir less obvious, so not sure
the output path for data_prep is the input path in the current approach. 

* If anything I would combine inputs and keep outputs separate. That the first argument is a single vector of paths is more user-friendly than having two separate arguments. This is why:


``` r
library(fs)
library(magrittr)

st_read_compute_and_write <- function(from_paths, to_path) { 
  # ... read from input_pahts
  # ... compute something
  message("Writing `to_path`: ", to_path)
  # ... write something
  
  invisible(from_paths)
}

parent <- tempdir()
input1 <- path(parent, "input1")
input2 <- path(parent, "input2")

parent %>% 
  dir_ls(recurse = TRUE) %>%  # All inputs are under the same parent directory
  st_read_compute_and_write(to_path = "/somewhere/")
#> Writing `to_path`: /somewhere/
```

<sup>Created on 2021-11-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

--

> the output path for data_prep is the input path in the current approach. This might be confusing.

This is a challenge we bring to ourselves needlessly. The problem goes away if we follow the principle that each function should do one thing and only one thing well. With this API the fact that the output of prep is the input of run is now expected:

```r
"/path/to/inputs/dir" %>%
  fs::dir_ls(recurse = TRUE) %>%
  st_read() %>%
  st_prep() %>%
  st_run() %>%
  st_write("/path/to/output/dir")
```

--

> it is becoming increasingly painful to change the interface due to the needed adjustments in readme, vignette, tests etc, we should keep an eye on that.

This is a curse now but a blessing later, because README and the vignettes becomes our integration tests. I hear you now it's a pain and maybe I would maintain only a bare bones README.

PR #200, opened 2021-11-12
add error

MirijaHa:
added proper check to avoid confusing current user info 
```Error in gzfile(file, mode) : cannot open the connection                                                                                                                     
In addition: Warning message:
In gzfile(file, mode) :
 
 Error in gzfile(file, mode) : cannot open the connection ```


jacobvjk (approved):
lgtm

PR #208, opened 2021-11-15
infromation error

MirijaHa:
This PR makes sure that an informative error msg is thrown if obligarory input args are not provided. 

@maurolepore 
I am not too happy with this implementation as it takes a detour. I.e. it checks if arguments in the captured input arguments lists are symbols to infer their missingness.

I know of the arg = NULL, arg %||% stop(msg) pattern but would rather not use this here, as it will for sure cause confusion when users are informed of the argument defaults as planned here: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/189

Do you have an alternative idea? If not we can also merge like this as we need it by tmrw and it gets the job done 🤷


maurolepore (approved):
LGTM.

> Do you have an alternative idea? If not we can also merge like this as we need it by tmrw and it gets the job done shrug

If the user-facing behaviour is okay, then I suggest you merge to meet tomorrow's goal and leave a reminder (to revisit this) in your wish list for important but not urgent issues.

Here are the first thoughts that come to mind:

1. What error do we see by default? Do we need a better one?
2. Would it be easier to check for missing arguments in `run_stress_test_impl()`?
3. Maybe the `parent` argument of `abort` can help catch the low-level error and re-throw a nicer one? See the example's section at https://rlang.r-lib.org/reference/abort.html. I guess the default error would throw in `run_stress_test_impl()`, then we could catch it higher in the stack (closer to the user-facing function) and re-throw  a nicer one.

Right now these are questions to myself. No need to respond. It's more of a seed for re-gaining the context I now have in my head.

PR #209, opened 2021-11-15
mvp

MirijaHa:
logs and results for identical iter_args will overwrite each other, so we need to add timestamps.
In order to make this less fiddly and not explode file names even further I thought why not simply write them to subdirs in the output paths that are named after the timestamp.
Running change checks gets more annoying as one has to provide sub dir names. However since we are not going to change functionality much the coming days an we will integrate change_checks in stress testing framework soon I find it tolerable for now.


maurolepore:
My thoughts:

1. Can we avoid logging at all? Maybe by
    1.1 writing messages that the user can suppress, and
    1.2. spliting `run_stress_test()` into `run_stress_test_result_type_1()`, `*_type_2()`, and so on. With this approach, the information we need to communicate goes down to 1/5 (as we now spit 5 results at once), which makes it a lot more easy to digest as a printed message to the console.

If no, 

2.1. Can we reuse other logging frameworks? Here just a few hits:

``` r
pkgsearch::pkg_search("log")
#> - "log" --------------------------------------- 376 packages in 0.011 seconds -
#>   #     package       version  by                    @ title                   
#>   1 100 nnet          7.3.16   Brian Ripley         7M Feed-Forward Neural N...
#>   2  72 expm          0.999.6  Martin Maechler     10M Matrix Exponential, L...
#>   3  55 rappdirs      0.3.3    Hadley Wickham      10M Application Directori...
#>   4  55 futile.logger 1.4.3    Brian Lee Yung Rowe  5y A Logging Utility for R 
#>   5  42 lgr           0.4.3    Stefan Fleck         2M A Fully Featured Logg...
#>   6  37 logging       0.10.108 Walerian Sokolowski  2y R Logging Package       
#>   7  24 logger        0.2.1    Gergely Daróczi      4M A Lightweight, Modern...
#>   8  19 cranlogs      2.1.1    Gábor Csárdi         3y Download Logs from th...
#>   9  18 log4r         0.3.2    Aaron Jacobs         2y A Fast and Lightweigh...
#>  10  18 plogr         0.2.0    Kirill Müller        4y The 'plog' C++ Loggin...
```

If no,

* Can we save to temporary files, which never overwrite themselves

``` r
identical(tempfile(), tempfile())
#> [1] FALSE
```

MirijaHa:
> My thoughts:
> 
> 1. Can we avoid logging at all? Maybe by
>    1.1 writing messages that the user can suppress, and
>    1.2. spliting `run_stress_test()` into `run_stress_test_result_type_1()`, `*_type_2()`, and so on. With this approach, the information we need to communicate goes down to 1/5 (as we now spit 5 results at once), which makes it a lot more easy to digest as a printed message to the console. This is in line with the best practice that each function should do only one thing. If we can't row-bind the 5 results without confusing the user, then we should probabbly not spit 5 results in one run.
> 
> If no,
> 
> 2.1. Can we reuse other logging frameworks? Here just a few hits:
> 
> ```r
> pkgsearch::pkg_search("log")
> #> - "log" --------------------------------------- 376 packages in 0.011 seconds -
> #>   #     package       version  by                    @ title                   
> #>   1 100 nnet          7.3.16   Brian Ripley         7M Feed-Forward Neural N...
> #>   2  72 expm          0.999.6  Martin Maechler     10M Matrix Exponential, L...
> #>   3  55 rappdirs      0.3.3    Hadley Wickham      10M Application Directori...
> #>   4  55 futile.logger 1.4.3    Brian Lee Yung Rowe  5y A Logging Utility for R 
> #>   5  42 lgr           0.4.3    Stefan Fleck         2M A Fully Featured Logg...
> #>   6  37 logging       0.10.108 Walerian Sokolowski  2y R Logging Package       
> #>   7  24 logger        0.2.1    Gergely Daróczi      4M A Lightweight, Modern...
> #>   8  19 cranlogs      2.1.1    Gábor Csárdi         3y Download Logs from th...
> #>   9  18 log4r         0.3.2    Aaron Jacobs         2y A Fast and Lightweigh...
> #>  10  18 plogr         0.2.0    Kirill Müller        4y The 'plog' C++ Loggin...
> ```
> 
> If no,
> 
> * Can we save to temporary files, which never overwrite themselves
> 
> ```r
> identical(tempfile(), tempfile())
> #> [1] FALSE
> ```

2. since we need this tomorrow I would prefer to add that to the wishlist

1.1. I do not think we should give the user an option to suppress the log output to the files, it is relevant information that is needed to understand the analysis. The only information we could in theory allow to suppress are the 5 lines generated by cat, however I am not sure if it is worth giving the option since this would add 1 more argument to the interface. If you are strongly annoyed by them though we can add a switch via an env var that is available for developers.

1.2 I do not knowhow the results are generated contentwise, but many results build upon each other, so I think they cannot be split. Maximum a discrimination between market risk results and credit risk results could be made. 
However i do not understand how this relates to the log output? The information we write to the log files depends on the inputs, especially the PACTA results and not the result that is being calculated, so splitting the runs between the options would instead multiply the information that we need to report.

maurolepore:
>  The information we write to the log files depends on the inputs, especially the PACTA results and not the result that is being calculated

Interesting! If the side-effect of printing information to the console is separate from the computation, can the user get a function that does just the printing and another one that does just the computation?

```r
paths %>% 
  print_inputs_report() %>%  # Now `cat()` is the right function because the main goal is to print something
  run_stress_test() %>%  # The main goal is to compute something. Any side-effects uses `message()`, `warning()` or `stop()` (or friends).  
```


Any violation of good design principles suggest the current design  could improve.

https://design.tidyverse.org/side-effect-soup.html

MirijaHa:
> > The information we write to the log files depends on the inputs, especially the PACTA results and not the result that is being calculated
> 
> Interesting! If the side-effect of printing information to the console is separate from the computation, can the user get a function that does just the printing and another one that does just the computation?
> 
> ```r
> paths %>% 
>   print_inputs_report() %>%  # Now `cat()` is the right function because the main goal is to print something
>   run_stress_test() %>%  # The main goal is to compute something. Any side-effects uses `message()`, `warning()` or `stop()` (or friends).  
> ```
> 
> Any violation of good design principles suggest the current design could improve.
> 
> https://design.tidyverse.org/side-effect-soup.html

To some extend yes -> wishlist

jacobvjk (approved):
lgtm

PR #212, opened 2021-11-15
New data/ stress_test_arguments

maurolepore:


MirijaHa (approved):
Very nice, also agree that the new naming improves this further.

PR #215, opened 2021-11-15
(Minor) Multiple inputs to asset_type throws surprising error message

maurolepore:
Overall I think this is minor.

This PR exposes what I think is unespected behaviour: I expect an error about `asset_type` before an error about missing arguments.

``` r

too_long <- 1:2
run_stress_test(asset_type = too_long)
#> -- Running transition risk stress test
#> Error: All input arguments need to hold values.
#> x Missing arguments: input_path_project_specific, input_path_project_agnostic, output_path.
#> ℹ Did you provide all input arguemnts correctly?.
```

<sup>Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

-

The error I expect is this:

``` r

too_long <- 1:2
run_stress_test(asset_type = too_long, "a", "b", tempdir())
#> -- Running transition risk stress test
#> Error: Must not provide more than one value for not iterateable argument
#> x Arguments with multiple values: asset_type.
#> ℹ Please coorect your function call
```

<sup>Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/tests/testthat/_snaps/run_stress_test.md
+++ b/tests/testthat/_snaps/run_stress_test.md
@@ -1,0 +1,6 @@
+# with multiple inputs to non-iterable arguments errors gracefully
+
+    All input arguments need to hold values.
+    x Missing arguments: input_path_project_specific, input_path_project_agnostic, output_path.
+    i Did you provide all input arguemnts correctly?.
+
--- a/tests/testthat/_snaps/run_stress_test.md
+++ b/tests/testthat/_snaps/run_stress_test.md
@@ -1,5 +1,11 @@
 # with multiple inputs to non-iterable arguments errors gracefully
 
+    Must not provide more than one value for not iterateable argument
+    x Arguments with multiple values: asset_type.
+    i Please coorect your function call
+
+---
+
     All input arguments need to hold values.
     x Missing arguments: input_path_project_specific, input_path_project_agnostic, output_path.
     i Did you provide all input arguemnts correctly?.
--- a/tests/testthat/test-run_stress_test.R
+++ b/tests/testthat/test-run_stress_test.R
@@ -21,3 +21,9 @@ test_that("output includes argument names with suffix '_arg'", {
   cols_from_arguments_have_suffix_arg <- purrr::map_lgl(data, ~rlang::has_name(.x, "term_arg"))
   expect_true(all(cols_from_arguments_have_suffix_arg))
 })
+
+test_that("with multiple inputs to non-iterable arguments errors gracefully", {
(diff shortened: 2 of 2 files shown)


tests/testthat/_snaps/run_stress_test.md
```diff
@@ -0,0 +1,6 @@
+# with multiple inputs to non-iterable arguments errors gracefully
+
+    All input arguments need to hold values.
+    x Missing arguments: input_path_project_specific, input_path_project_agnostic, output_path.
+    i Did you provide all input arguemnts correctly?.
```
maurolepore, line 5: FIXME: This error message is not what we want if we run `run_stress_test(asset_type = 1:2)`.

This highlights the need for automated tests ASAP.

MirijaHa:
> Overall I think this is minor.
> 
> This PR exposes what I think is unespected behaviour: I expect an error about `asset_type` before an error about missing arguments.
> 
> ```r
> library(r2dii.climate.stress.test)
> 
> too_long <- 1:2
> run_stress_test(asset_type = too_long)
> #> -- Running transition risk stress test
> #> Error: All input arguments need to hold values.
> #> x Missing arguments: input_path_project_specific, input_path_project_agnostic, output_path.
> #> ℹ Did you provide all input arguemnts correctly?.
> ```
> 
> Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)
> 
> 
> The error I expect is this:
> 
> ```r
> library(r2dii.climate.stress.test)
> 
> too_long <- 1:2
> run_stress_test(asset_type = too_long, "a", "b", tempdir())
> #> -- Running transition risk stress test
> #> Error: Must not provide more than one value for not iterateable argument
> #> x Arguments with multiple values: asset_type.
> #> ℹ Please coorect your function call
> ```
> 
> Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)

Can u elaborate why you would expect that order of errors? I would expect it the other way round as it is currently implemented. The rationale behind that is to:

-  first: look  at the overall pattern (i.e. are all arguments provided)
- second: Look at details of single arguments

maurolepore:
In general I expect the first argument to be checked first. That way as I user I can learn how to use the tool just by trial and error.

In this case you're right that missing arguments should be checked before, but that would also meet my expectation if this function was designed following the most common pattern -- where required arguments come before details arguments (with defaults).

So you revealed that my surprise comes ultimately because as a user I learned to expect a pattern that this function violates. 

https://design.tidyverse.org/args-data-details.html

PR #218, opened 2021-11-16
Bye bye lookups

MirijaHa:
This PR:

- removes functionality using old lookups
- adds nicer rlang style error messages if wrong inputs are provided
NOTE: this should be on master before today at 5 🙏


jacobvjk (approved):
nice one!
lgtm

PR #222, opened 2021-11-17
Return option

MirijaHa:
This PR adds the option to return st results in R instead of exporting them to the output directory. @jakubcervenka and @daisy-pacheco  might like this.

NOTE: I merely copy pasted code around and did not refactor any internals to gain implementation speed.

@jacobvjk 
I did not include the new option in the vignettes yet, would be great if you could do it in a follow up.

On change checks:

- checks passed for all asset types with and without iteration
- I compared also results that are returned vs exported and imported. There are some detail differences on data types (glue vs char, numeric vs integer), but I would say this should not be a problem.


R/run_stress_test.R
```diff
@@ -4,7 +4,10 @@
 #' understand sensitivities of the scenarios, in which case the user may pass a
 #' vector of values to one (and only one) of the detail arguments. This will
 #' result in running the analysis multiple times in a row with the argument
-#' varied. NOTE: argument `asset_type` is not iterateable.
+#' varied.
+#' NOTE: argument `asset_type` is not iterateable.
+#' NOTE: if `return_results` is TRUE results will not be written to `output
```
jacobvjk, line 9: Does this mean that output path becomes optional? Do we check if at least one is given? Or do we need both arguments regardless? Maybe something we need to be extra clear about  with the users

R/write_results.R
```diff
@@ -1,127 +1,40 @@
...
-    ) %>%
-    dplyr::ungroup() %>%
-    dplyr::arrange(
-      .data$scenario_geography, .data$scenario_name, .data$investor_name,
-      .data$portfolio_name, .data$ald_sector, .data$year
-    )
+  results_list$stress_test_results_comp_el %>%
+    readr::write_csv(file.path(
+      output_path,
+      paste0("stress_test_results_", asset_type, "_comp_el_", iter_var, ".csv")
```
jacobvjk, line 28: why not always use `glue::glue()` for consistency?

jacobvjk (commented):
some small comments

R/write_results.R
```diff
@@ -1,127 +1,40 @@
...
-    report_all_duplicate_kinds(
-      composite_unique_cols = c(
-        "scenario_name", "scenario_geography", "investor_name", "portfolio_name",
-        "ald_sector", "term", sensitivity_analysis_vars
-      )
-    ) %>%
+  results_list$stress_test_results_sector_pd_changes_overall %>%
     readr::write_csv(file.path(
       output_path,
       paste0("stress_test_results_", asset_type, "_sector_pd_changes_overall_", iter_var, ".csv")
```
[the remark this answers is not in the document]
MirijaHa: done

R/write_results.R
```diff
@@ -1,127 +1,40 @@
...
-    report_all_duplicate_kinds(
-      composite_unique_cols = c(
-        "scenario_name", "scenario_geography", "investor_name", "portfolio_name",
-        "ald_sector", "year", sensitivity_analysis_vars
-      )
-    ) %>%
+  results_list$stress_test_results_sector_pd_changes_annual %>%
     readr::write_csv(file.path(
       output_path,
       paste0("stress_test_results_", asset_type, "_sector_pd_changes_annual_", iter_var, ".csv")
```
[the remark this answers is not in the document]
MirijaHa: done

R/write_results.R
MirijaHa: done

R/run_stress_test.R
MirijaHa: if return_results is TRUE output path does not need to be provided. Adjsuting.

MirijaHa: actually,  I have to correct myself. even if result are returned we write log files to output paths, so path needs to be provided. Guess that couls be emphasised in vignette.

jacobvjk: okay

jacobvjk (approved):
lgtm

maurolepore (commented):
Note it's generally best to aim for a function to be type stable 

https://vctrs.r-lib.org/articles/stability.html

PR #224, opened 2021-11-17
dummy log entry

MirijaHa:
make sure there is always sth in the output subdir, even if st fails.
Inspired by @daisy-pacheco


jacobvjk (approved):
lgtm

PR #225, opened 2021-11-17
Final update vignette malaysia

jacobvjk:
explanation of the `return_results` option

### DIFF
--- a/vignettes/articles/03-run-stress-test.Rmd
+++ b/vignettes/articles/03-run-stress-test.Rmd
@@ -66,13 +66,14 @@ log file into the designated output directory, using the default values for all
 detail arguments.
 
 **NOTE:** The default mode will **not return** any objects within the R session.
-We find this is the easiest and most user friendly way to obtain results, but we
-acknowledge that more advanced R users may find this to depart from the
-conventions of other packages. Below we highlight, how to return the results as
-objects within the R session.
+We find this is the easiest and most user friendly way to obtain results. We
+alternatively provide the option that the function returns a list holding the
+result objects. This allows the user to inspect and further analyze the data
+directly in R.
+
 
 ### Input arguments
-Specifications on the detail arguments are summarised in the following overview.
+Specifications on the detail arguments are summarized in the following overview.
 The defaults that are used if the argument is not adjusted when calling
 `run_stress_test()` can be found in column `default`. The user may change any
 detail argument by explicitly passing it to the function. Note that there are
@@ -114,8 +115,9 @@ dedicated mechanism in place that handles companies which are phasing out
 certain technologies within the time frame of the production forecast.
 * `return_results` Logical parameter that indicates if the outputs of the
 analysis should be directly written to the indicated output path or simply
-returned as objects within the running session. The default is FALSE, which
-writes the outputs to the target directory and does not return them as objects.
+returned as a list of results within the running session. The default is FALSE,
+which writes the outputs to the target directory and does not return them as a
+list object in R.
 
 ### Adjusting the input arguments
 The user may want to run a stress test that is tailored to their assumptions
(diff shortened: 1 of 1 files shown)


vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -65,6 +65,12 @@ This will calculate the stress test for loans and write the results along with a
 log file into the designated output directory, using the default values for all
 detail arguments.
 
+**NOTE:** The default mode will **not return** any objects within the R session.
+We find this is the easiest and most user friendly way to obtain results, but we
+acknowledge that more advanced R users may find this to depart from the
+conventions of other packages. Below we highlight, how to return the results as
```
MirijaHa, line 71: I would not say convention of other packages, this phrase does not make sense and I could banter on on why, instead I would suggest to say sth like>
We alternatively provide the option that the function returns a list holding the result objects. This allows the user to inspect and further analyse the data directly in R.

vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -106,6 +112,10 @@ company and technology should be excluded from the analysis. At this point, it
 is strictly recommended to set this to TRUE (default) because there is not yet a
 dedicated mechanism in place that handles companies which are phasing out
 certain technologies within the time frame of the production forecast.
+* `return_results` Logical parameter that indicates if the outputs of the
+analysis should be directly written to the indicated output path or simply
+returned as objects within the running session. The default is FALSE, which
```
MirijaHa, line 117: - list of results.
- I was wondering if to add a use example here, but then again who wants to use this option should be able to understand it without help.

vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -124,15 +134,27 @@ run_stress_test(
 )
 ```
 
+**NOTE:** One important input argument that affects the behavior of where the
+results are written, is the argument `return_results`. Per default it is set to
+`FALSE`, which means that results are written as csv files to a sub directory in
+the `output_path` and the user must open these files to view the results.
+If the user wants to view the results directly in RStudio, they can set the
```
MirijaHa, line 141: R, not RStudio

vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -124,15 +134,27 @@ run_stress_test(
...
+argument to `TRUE` which will return them as data frames in the R session. This
```
MirijaHa, line 142: list of dataframe

vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -124,15 +134,27 @@ run_stress_test(
...
+means they are not written to the designated `output_path` however, so the user
```
MirijaHa, line 143: if themselves then users

MirijaHa (commented):
nits

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: Thinking about it, I believe we should consider a follow up (non blocking for the next week) to return the results as 5 separate objects. If we aim at user friendliness and expect that some users will not have used R much, I think a list type object still presents somewhat of a barrier.

MirijaHa: a list is the only way to cleanly return >1 objects 🤡  , unless we adjust the function in a way that it only returns 1 result type per run. 
I would say being able to deal with a list is a nicer selector for if one should use the option

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: I wanted to use a gender neutral they here. Does this imply I should use plural? I thought this worked different..

MirijaHa: ahh, i thought I deleted this comment, your formulation is correct afaik

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: fine by me, let's see what the users think

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: done

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: done

vignettes/articles/03-run-stress-test.Rmd
jacobvjk: done (and no)

PR #230, opened 2021-11-18
No more term val arg

MirijaHa:
As @jakubcervenka  found out that altering terminal_value has currently no effect on the model we temporarily do not expose it as arg as argument.
I removed related functionality as far as I could without creating problems fro webtool.


jacobvjk (approved):
beautiful

data-raw/stress_test_arguments.R
```diff
@@ -4,7 +4,7 @@ stress_test_arguments <- tibble::tribble(
                "asset_type", "character",       NA, "equity, bonds, loans",         NA,         NA,
         "lgd_senior_claims",    "double",   "0.45",                     NA,      "0.3",      "0.6",
   "lgd_subordinated_claims",    "double",   "0.75",                     NA,      "0.6",      "0.9",
-           "terminal_value",    "double",      "0",                     NA,        "0",      "0.1",
+#           "terminal_value",    "double",      "0",                     NA,        "0",      "0.1",
```
maurolepore, line 7: Cool. Shall I delete this line? I'll submit a follow up updating the snapshot of this dataset. I could that both at once.

tests/testthat/_snaps/stress_test_arguments.md
```diff
@@ -3,17 +3,16 @@
...
+        <chr>                   <chr>     <chr>   <chr>                <chr> <chr>
+      1 asset_type              character <NA>    equity, bonds, loans <NA>  <NA> 
+      2 lgd_senior_claims       double    0.45    <NA>                 0.3   0.6  
+      3 lgd_subordinated_claims double    0.75    <NA>                 0.6   0.9  
+      4 risk_free_rate          double    0.02    <NA>                 0     0.05 
+      5 discount_rate           double    0.02    <NA>                 -0.01 0.05 
+      6 div_netprofit_prop_coef double    1       <NA>                 0.8   1    
+      7 shock_year              double    2030    <NA>                 2025  2035 
+      8 term                    double    2       <NA>                 1     10   
+      9 company_exclusion       logical   TRUE    TRUE, FALSE          <NA>  <NA>
```
maurolepore, line 17: Oh cool, you updated the snapshot!!! Are you manually changing the `is_me()` to register yourself? Now that we can write outputs outside dropbox, or return them as a list, then I think I could adapt the tests to they automatically run for you, me, and whoever runs `devtools::test()`.

data-raw/stress_test_arguments.R
MirijaHa: i updated the snap I think when prompted to? Or do you see it being out of synch?
I would say the line can stay for now, we will likely bring it back in a few days..

tests/testthat/_snaps/stress_test_arguments.md
MirijaHa: Ah you noticed, so all good.
I did not change is_me, this snapshot test should not be wrapped in it.

Before adapting the current tests let us please talk about it. I would like to use the opportunity to automated the change_checks and extend covered cases.

data-raw/stress_test_arguments.R
maurolepore: Ah, you did well! I thought I was running that test for myself only but obviously not (and it was intentional because it's safe to run for anyone).

A closer look shows this test does not use `is_me()`.

```r
test_that("hasn't changed", {
  expect_snapshot(stress_test_arguments)
})
```

PR #233, opened 2021-11-22
Ado 3007 rm

MirijaHa:
As agreed removed the outdata workflow external_stress_test_on_existing_results.R
Using the script to find unused functions I then removed functions that are no longer needed.


jacobvjk (approved):
great one!

PR #234, opened 2021-11-22
Ado 3008 uncalled

MirijaHa:
This PR removes file global_functions from root level.
All functions where either obsolete or only called from 1 script, in which case I move the function definition to the respective scripts.
Chances are in a follow up I can get rid of portfolio_input_check_functions too.


jacobvjk (approved):
lgtm

Issue #235, opened 2021-11-22
`set_params_st.R` needed for webtools


cjyetman:
@MirijaHa #233 breaks the web tools my removing `R/set_params_st.R`, which is sourced here:
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/05c9d2ec313a0f31c132c0f36187f1c6d5a3c9ba/web_tool_external_stress_test.R#L27
and here:
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/05c9d2ec313a0f31c132c0f36187f1c6d5a3c9ba/web_tool_stress_test.R#L43

MirijaHa:
Thanks for noticing.
The function that is defined in the file `set_project_parameters` is not called from web_tool_external_stress_test.R and web_tool_stress_test.R, which is why I did not notice. I will open a follow up that removes the lines you posted.

PR #236, opened 2021-11-22
rm refs to set_params_st

MirijaHa:
fixes https://github.com/2DegreesInvesting/r2dii.climate.stress.test/issues/235


cjyetman (approved):
thanks! 🙏 

to be honest, there may be similar files that are sourced here and were deleted in #233 that I haven't noticed yet because it already errored out at `set_params_st.R`, so it might be prudent to compare the list of files deleted in #233 to the list of files sourced here, but I leave that up to you.

PR #238, opened 2021-11-23
Move yaml

MirijaHa:
As another decoupling prep moved yml fall to nonstandard dir since it is only needed there


jacobvjk (approved):
lgtm

PR #239, opened 2021-11-23
Export rampage

MirijaHa:
This PR removes unneeded export to simplify orientation.
I kept export for all visualisation functions as we can probably make them useful to the user, some util functions where I was not sure, and all functions called directly from the webtool (although I think this is not needed, since files are sourced, but I would rather disentangle that in a follow up).


jacobvjk (approved):
lgtm

PR #240, opened 2021-11-24
Add test to check calc_survival_probability_merton output

rht:


rht:
@MirijaHa closing a pull request without giving any explanation is not cool. Do you mind to explain why this PR was closed?

jacobvjk:
Dear @rht, thanks for pointing to that. At this point, we still have many fundamental tasks to solve, so uncoordinated PRs will produce more work than we currently feel we can handle. As this is still a beta version many changes are still to be expected and as such, we would ask for coordination before we currently approve any external PRs. In this specific case, we cleaned up old issues and PRs in the repository and accidentally missed to add an explanation. We have previously discussed this with Moritz Bär internally, so we assumed this was clear. Have a great week and more than open to your help and collaboration in case we coordinate targets and next steps more extensively in the future. Best, Jacob

rht:
OK, understandable. Thank you for the explanation! With this, I can sleep better.

PR #241, opened 2021-11-24
split functions

MirijaHa:
This PR splits webtool and standard version for read_capacity_factors to simplify decoupling.
Change checks pass, also I compared that imported old and new cap data do not differ between master an PR branch


jacobvjk (approved):
lgtm

PR #242, opened 2021-11-24
Ado 3057 fin

MirijaHa:
Another disentaglement, this time for old and new version of pricedata. I know we had originally agreed on the current interface, but as for user stress testing we will move to the new price data soon this helps for disentanglement.
The checking for technologies should better not be part of the function but I did not want to risk any webtool problem.

Change checks passed, also I compared price data imported with master and dev version.

FYI, the function told me that both for price_data_long_2021Q3.csv and price_data_long_2021Q1.csv the 
 technologies: Electric" "Hybrid"   "ICE"  are missing, is this expected? also tagging @jakubcervenka  because i think he worked on the data too https://github.com/2DegreesInvesting/r2dii.stress.test.data/commit/1cf5f225707fd458eb71922e2c77d1f84d522ec5


jacobvjk:
@MirijaHa @jakubcervenka the prices for automotive should be unit prices, i.e. 1

MirijaHa:
> @MirijaHa @jakubcervenka the prices for automotive should be unit prices, i.e. 1

thx for the info, will create.a ticket for @jakubcervenka

jacobvjk (approved):
lgtm

PR #243, opened 2021-11-25
validate in main

MirijaHa:
This PR is a preparation for https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3068 in which I will extract the data import out of run_stress_test_impl.

This PR runs validation of input args in the mapper, not in run_stress_test_impl, which means validation has to be able to deal with long vectors.


jacobvjk (approved):
we are reaching a stage where it becomes more difficult for me to review that every altered piece makes sense. I get the point of this and it looks right to me, but without spending much time on a detailed review I will have to simply use that to say
LGTM

PR #250, opened 2021-12-01
Run usethis::use_tidy_style()

maurolepore:
Loosely relates to [AB#3100](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/3100)

Note this works on R/, tests/, and data-raw/.

It's worth inspecting the diff because some changes
may be undesirable, for example, the readable layout
of tibble::tribble() would be broken -- to avoid that
we can use "# styler: off" and "# styler: on" around
the code we won't to omit.

﻿

### DIFF
--- a/R/lookup.R
+++ b/R/lookup.R
@@ -59,6 +59,7 @@ equity_market_filter_lookup <- "GlobalMarket"
 
 # technology and sector mapping between P4I and P4B
 # HDV and shipping not consistently defined across both versions at this time
+# styler: off
 p4i_p4b_sector_technology_lookup <- tibble::tribble(
   ~sector_p4b,   ~technology_p4b,             ~sector_p4i,    ~technology_p4i,
   "automotive",  "electric",                 "Automotive",   "Electric",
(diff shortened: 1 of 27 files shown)


MirijaHa (approved):
stylish

data-raw/stress_test_arguments.R
```diff
@@ -1,16 +1,16 @@
 # style: off
 stress_test_arguments <- tibble::tribble(
-                      ~name,       ~type, ~default,               ~allowed,       ~min,       ~max,
```
maurolepore, line 3: I'll undo this change.

R/lookup.R
```diff
@@ -60,47 +60,49 @@ equity_market_filter_lookup <- "GlobalMarket"
 # technology and sector mapping between P4I and P4B
 # HDV and shipping not consistently defined across both versions at this time
 p4i_p4b_sector_technology_lookup <- tibble::tribble(
-  ~sector_p4b,   ~technology_p4b,             ~sector_p4i,    ~technology_p4i,
```
maurolepore, line 63: I'll undo this change

MirijaHa (approved):
thx for fixing the tibbles

lorenzwalthert:
@maurolepore tribble alignment (and alignment in general) can be preserved under certain circumstances, see https://styler.r-lib.org/articles/detect-alignment.html#examples.

PR #252, opened 2021-12-01
Test integration with the webtool on CI

maurolepore:
Relates to [AB#3137](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/3137)

This PR runs the webtool scripts on every commit to this repo. Any
breaking change should make this workflow fail, for example:

* 38a9110 breaks web_tool_stress_test.R and this workflow [fails](https://github.com/2DegreesInvesting/r2dii.climate.stress.test/runs/4387106874?check_suite_focus=true).

* e249dc4 breaks web_tool_external_stress_test.R and this workflow [fails](https://github.com/2DegreesInvesting/r2dii.climate.stress.test/runs/4387232974?check_suite_focus=true).

--

Note the workflow is identical to that in PACTA_analysis but the behaviour is different because the repo that triggers the workflow is different. I'll highlight a comment that (tersely) explains this point.


.github/workflows/source-web-tool-scripts.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: Source web tool scripts in container r2dii/r-packages
+    strategy:
+      fail-fast: false
+    env:
+      R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      # When checking out the repository that triggered a workflow, `ref:`
+      # defaults to the reference or SHA for that event. Otherwise, uses the
+      # default branch --https://github.com/actions/checkout
```
maurolepore, line 16: Here the repo that triggers the workflow is r2dii.climate.stress.test. So the workflow runs with the last commit on this PR and the default branch of all other repos. That means the same workflow is used with different intentions here and in PACTA_analysis. Here it helps us notice if any change here breaks the webtool. In PACTA_analysis it helps notice if any change in PACTA_analysis breaks the webtool -- and as a side effect it helps detect changes in the friends of PACTA_analysis that have breaking changes in the default branch. This of course would not happen if all the friends of PACTA_analysis did what we are doing here -- i.e. to avoid merging into the default branch any PR that breaks the web tool.

maurolepore:
Approved: AB#3137

Test if I can change the state of a ticket from a pull request.

maurolepore:
Done: AB#3137

Test if I can change the state of a ticket from a pull request.

maurolepore:
Done: #3137

Test if I can change the state of a ticket from a pull request.

cjyetman:
@maurolepore have you notice these messages?
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/runs/4387289792?check_suite_focus=true#step:3:25
```
(node:18) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```
Does that mean we can expect that there will be a new version of actions/checkout soon that we will need to change this to since we explicitly use `actions/checkout@v2`?

I'm also setting this up in `create_interactive_report`, plus I want to keep up to date on necessary changes in PACTA_analysis.... that's why I'm asking.

cjyetman:
No, I don't ☹️. But it's good to have it fresh in our minds that a change is probably coming eventually.

I also just now noticed that `actions/checkout@v2` actually has [many minor version updates](https://github.com/actions/checkout/releases) (currently at [v2.4.0](https://github.com/actions/checkout/releases/tag/v2.4.0)), so maybe whatever fix that they might implement for this message we might automatically benefit from anyway?

maurolepore:
Good to know they released a new minor version.

In my experience so far, keeping up to date with the workflows that r-lib provides has taken relatively little effort. If you experiment with the latest version it'd be good to know what you find.

cjyetman (approved):
love it!

MirijaHa (approved):
what CJ said

PR #246, opened 2021-11-29
Update pr-commands.yaml workflow

maurolepore:
Relates to [AB#3100](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/3100)

This PR updates a workflow that allows us to run an automated styler on a PR. That means the contributor doesn't need to know how to do that locally.

### `?use_github_action_pr_commands()`

> This workflow enables the use of two R-specific commands in pull request issue comments:
* `/document` to run `roxygen2::roxygenise()` and update the PR
* `/style` to run `styler::style_pkg()` and update the PR

PR #263, opened 2021-12-07
Update pr-commands.yaml workflow

maurolepore:
Fixes #246 -- which accidentaly touched more files than expected.

For details see #246.﻿

### DIFF
--- a/.github/workflows/pr-commands.yaml
+++ b/.github/workflows/pr-commands.yaml
@@ -1,55 +1,75 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   issue_comment:
     types: [created]
+
 name: Commands
+
 jobs:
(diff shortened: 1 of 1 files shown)


MirijaHa (commented):
I am not sure i totally understand this PR and want to make sure I understand it. As far as I see this PR only changes:

maurolepore:
> Does this mean we actually already have a job to potentially do styling and this PR jsut updates the workflow to it`s newest version?

Exactly. We already have [.github/workflows/pr-commands.yaml](https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/master/.github/workflows/pr-commands.yaml). When I run `usethis::use_github_action_pr_commnands()` the latest version of that file overwrites the existing one. I didn't look at the diff. I just trust the author of the workflow has a reason for the update. 

My motivation for updating was that the command `/style` seemed to not work, so before exploring any further I first update to the latest version of the workflow.

MirijaHa (approved):
🥇

PR #264, opened 2021-12-08
Check to dos

jacobvjk:
This PR closes ADO 2392 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2392

- Outdated todos are removed
- ADO tickets are commented in, where it is clear when the todo will be resolved
- I did not touch non-user facing scripts such as checks
- litigation risk also still with many to dos. This should be refactored properly soon


MirijaHa (approved):
Pure happyness

PR #265, opened 2021-12-08
Prep 3122

MirijaHa:
Removing outdated functionality from webtool:
- company exclusion
- calculation level other than portfolio
- also first real live use case for @maurolepore `s integration test 👻


web_tool_external_stress_test.R
```diff
@@ -118,8 +120,7 @@ if (exists("portfolio")) {
     dplyr::mutate(
       loss = exposure * shock / 100,
       sector = ifelse(is.na(sector), "Other", sector)
-    ) %>%
-    dplyr::ungroup()
+    )
```
MirijaHa, line 123: just to go sure, since I am not good with groupings and cannot test the code, you moved the grouping from l 123 (after the mutate) to l 112 (before the mutate), this has no impact no the results? @jacobvjk

jacobvjk: yes, we are not doing column operations here, so this ought to make no difference, not sure why it was part of the grouping in the first place

jacobvjk (approved):
lgtm

PR #266, opened 2021-12-08
Ado 3122

MirijaHa:
Relates to [AB#3122](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/3122)

This PR removes unneeded exports. I am not sure what to do about the exported pipe operator, it comes from using utils::use_pipe I think, any opinion @maurolepore ?


jacobvjk (approved):
lgtm
but would prefer if @maurolepore could also check

maurolepore (approved):
LGTM. 

Before merging consider adding the tag `#' @noRd` wherever you see roxygen2 comments (`#'`) on a function that is not exported. This stops the generation of .Rd files.

The code below is convoluted because I'm adding `@noRd` programatically for the reprex. Doing it manually is a lot easier. This code shows that even though you no longer export the function `degrees()`, there is a file "man/degrees.Rd". After I add `#' @noRd` and redocument, the file is gone.

``` r
devtools::document()
#> ℹ Updating r2dii.climate.stress.test documentation
#> Writing NAMESPACE
#> Writing NAMESPACE
file.exists("man/degrees.Rd")
#> [1] TRUE

# Add @noRd to tell roxygen2 not to generate an .Rd file
no_rd <- "#' @noRd\n"
f <- "degrees <- function"
path <- "R/utils.R"
lines <- sub(f, paste0(no_rd, f), readLines(path))
writeLines(lines, path)
devtools::document()
#> ℹ Updating r2dii.climate.stress.test documentation
#> Writing NAMESPACE
#> Writing NAMESPACE
#> Writing get_st_data_path.Rd
#> Writing data_path.Rd
#> Writing dropbox_exists.Rd
#> Deleting degrees.Rd
file.exists("man/degrees.Rd")
#> [1] FALSE
```

<sup>Created on 2021-12-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore:
I think we could use the pipe internally and not export it. But it may not hurt to leave it as is -- so the pipe is available in examples and available to users after they attach this package with `library()`. From `?usethis::use_pipe()` I see the pipe is exported by default, but you can opt out with `?usethis::use_pipe(export = FALSE)`.

PR #268, opened 2021-12-10
Abstraction and harmonisation 1

MirijaHa:
This PR appears useless as is:
it moves all reading in of data and the related wrangling to run_stress_test_impl instead of hiding it away in functions.  I did this because it is only the first step to achieve a bigger goal namely:

- to streamline what data import, wrangling, filtering and checking does between datasets. I think we will largely benefit from that latest when we do the pacta uncoupling or add more up to date data sets.
- to extract data import form calculation function (only 1 import needed for sensitivity analysis, maybe first step to removing imports from the scope of the package).

As getting an overview is really difficult I did this PR as a frist step that simplifies getting respective overview

I ran change checks on all asset types, no problems.


jacobvjk (approved):
beautifully useless
lgtm

PR #271, opened 2021-12-14
new deault and lims discount rate

jacobvjk:
closes ADO 3113: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3113

This PR:
- updates the default discount rate
- updates the max allowed discount rate
- updates the tests to reflect that


jacobvjk:
snapshot test expectedly failed as this leads to different results in the profit calculations

MirijaHa (commented):
Are these changes permanent or temporarily for the IFC?
I noticed snapshot tests fail for bonds and equity but not for loans workflow, is that in line with your expections?

jacobvjk:
(1) they are supposed to be permanent. (2) no, I would expect all of them to change... the discount rate is used in the DCF model which informs the credit risk model. therefore it should be relevant to loans as well.. will have to investigate

MirijaHa:
> (1) they are supposed to be permanent. (2) no, I would expect all of them to change... the discount rate is used in the DCF model which informs the credit risk model. therefore it should be relevant to loans as well.. will have to investigate

1 cool
2 found issue (unchanged min value is used in snapshot tests for loans)

MirijaHa (approved):
stunning

PR #272, opened 2021-12-15
move check from import

MirijaHa:
micro PR in the context of streamlining data import and wragning. Removes wranging from data import for price data. made a duplicate  fucntion to not mess with webtool.


jacobvjk (approved):
mesmerising
lgtm

PR #273, opened 2021-12-15
Prep pacta res

MirijaHa:
another micro PR. I create a duplicate of read pacta results that only works for offline workflow. Prep for data stuff streamlining and uncoupling of CRAN package and repo.


jacobvjk (approved):
lgtm

PR #274, opened 2021-12-16
Streamlining

MirijaHa:
In this PR I merely copy pasted all processings and check that are done into 1 function per dataset. This way it is easier to see current unneeded steps and inconsistencies
Snapshot tests pass.

Remaining steps in this endeavour:

- remove all inconssitencies and superfluous information
- take import, and if possible wrangling, out of the loop


jacobvjk (commented):
integration check with web tool successful?

MirijaHa:
> integration check with web tool successful?

it wasnt at first, but the intergation check found an issues which I then fixed 🤓

jacobvjk:
> > integration check with web tool successful?
> 
> it wasnt at first, but the intergation check found an issues which I then fixed 🤓

amazing

jacobvjk (approved):
lgtm

PR #276, opened 2021-12-16
Rm superfluous steps

MirijaHa:
Last cleaning step before moving import out of loop:

- i cleaned some unnneded assginments and made sure that lookup vars are only used in toplevel function. This should make it easier to make the code more flexible to new iterable variables
- moved some duplicates
- delete an, IMO illogical, handling in price data. Please have a sharp look if I overlooked sth there 🙏


R/process_data.R
```diff
@@ -108,8 +114,6 @@ process_excluded_companies <- function(data, company_exclusion, technologies) {
 process_df_price <- function(data, technologies, sectors, start_year, end_year) {
   data_processed <- data %>%
     check_technology_availability(expected_technologies = technologies) %>%
-    dplyr::filter(year >= start_year) %>%
```
jacobvjk, line 111: is this handled by using an inner join later on @MirijaHa ?

R/process_data.R
```diff
@@ -129,11 +133,6 @@ process_scenario_data <- function(data, start_year, end_year, sectors, technolog
                                   scenario_geography_filter, scenarios_filter) {
   data_processed <- data %>%
     wrangle_scenario_data(start_year = start_year, end_year = end_year) %>%
-    dplyr::filter(
```
jacobvjk, line 132: is this taken care of when reading the data or why should we remove these?

jacobvjk (commented):
see comments

R/process_data.R
MirijaHa: The filtering by start year (and also end year) is also done in l 115, so I deleted this line 111.
Line 112 looks useless to me. It basically checks that data$year > start_year. Since we ensure that by filtering anyway I removed the line. Then again there probably was a reason why this function was even written, so maybe I am overlooking something? 🤔

R/process_data.R
MirijaHa: they are duplicates to lines 137 ff

R/process_data.R
jacobvjk: no seems fine then

R/process_data.R
jacobvjk: oh wow yes..

jacobvjk (approved):
lgtm

Issue #270, opened 2021-12-14
automotive only portfolio can not run stress testing


cjyetman:
https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3214

When a user's portfolio that resolves to having only Automotive relevant assets, stress testing fails because there will be only one scenario here (SDS)
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/42a283aa46c213b7cd55d393c06ae6e6a8a4e584/web_tool_stress_test.R#L289-L294

The allowable scenarios are defined in the model_parameters config file, e.g.
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/42a283aa46c213b7cd55d393c06ae6e6a8a4e584/model_parameters/model_parameters_PA2021NO.yml#L11-L14

Which is read in here and set to `scenarios_filter`
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/42a283aa46c213b7cd55d393c06ae6e6a8a4e584/web_tool_stress_test.R#L116-L134

and then applied as a filter here
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/42a283aa46c213b7cd55d393c06ae6e6a8a4e584/web_tool_stress_test.R#L263-L277

Here's a test portfolio
[TestPortfolio_Input.csv](https://github.com/2DegreesInvesting/r2dii.climate.stress.test/files/7712546/TestPortfolio_Input.csv)

MirijaHa:
Thanks for the detailed overview on the issue @cjyetman 

As it affects the webtool I would suggest you take it over @jacobvjk 

Some thoughts though: 
- I only allowed use of portfolios with only 1 sector for the user workflows in this PR, in order to not alter webtool behavior https://github.com/2DegreesInvesting/r2dii.climate.stress.test/commit/05ebe6aaee5bd38bd7353542b54658431f3a3e1d#diff-b6b6854a02ae177f8860f49654ff250c1554d021ae434b6efdbe99d00bdc6055R23 this PR can likely give some orientation though
- taking a step back, you said that in general including portfolios with only 1 sector does not make sense for stresstesting, how do you evaluate this in the webtool case?
- If there was only SDS scenario included there would be no baseline, so no shocking can happen. This should not be a consequence of only including 1 sector, is this description of the behavior correct in this regard?

jacobvjk:
[...]

PR #277, opened 2021-12-17
Npsrts

MirijaHa:
closes issue: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/issues/270

I nearly would have moved the filtering to the wrong spot, i.e. after dplyr::filter(!(scenario == "ETP2017_NPS" & ald_sector == "Power")). This would have meant that we would have gotten duplicate entries for some sectors and the workflow would have failed later on. I still think it is good if you look at this sharply once more @jacobvjk 

I also removed some outdated comment and added a simply check to avoid that in some weird situation NPS and NPSRTS are mixed.


cjyetman (approved):
lgtm.... if I understand correctly, it looks like this code will automatically adapt to either "NPS" or "NPSRTS", which I think is a very good idea 👍🏻

jacobvjk (changes requested):
I think this may still be problematic, @MirijaHa because your check, which makes sense in principle, should fail if we have WEO2020_NPS and ETP2017_NPSRTS in the sample. This is likely going to be the case though. In the end this check would have to be run by scenario source

jacobvjk:
OR we need to run this check by sector and use the scenario source/sector filtering from the user version.. which I guess is more complicated

MirijaHa:
> I think this may still be problematic, @MirijaHa because your check, which makes sense in principle, should fail if we have WEO2020_NPS and ETP2017_NPSRTS in the sample. This is likely going to be the case though. In the end this check would have to be run by scenario source

I wonder if to simply remove this check, as actually based on the scenarioAnalysis files CJ provided it is actually impossible for NPS and NPSRTS to be in the same results, so it might be overly cautious to check this, what do you think?

jacobvjk:
> > I think this may still be problematic, @MirijaHa because your check, which makes sense in principle, should fail if we have WEO2020_NPS and ETP2017_NPSRTS in the sample. This is likely going to be the case though. In the end this check would have to be run by scenario source
> 
> I wonder if to simply remove this check, as actually based on the scenarioAnalysis files CJ provided it is actually impossible for NPS and NPSRTS to be in the same results, so it might be overly cautious to check this, what do you think?

while we are at it, maybe it will pass with your test, as we seem to get WEO2019_SPS and then change that to WEO219_NPS later on after the NPSRTS check. But you will notice that this is really starting to get out of control slowly. We are also not handling WEO2020_SPS, if that should be required at some point. So I can approve this, but we should overhaul the handling of this not too far into the future

jacobvjk:
we can also remove the check, you are right

jacobvjk (approved):
lgtm

PR #280, opened 2021-12-21
Flexible term pre

MirijaHa:
This is a prep PR for allowing input of individual loans.
I open it as a separate PR beforehand so it is possible to keep an overview of how it impacts results.
Here snapshot tests for bonds and equity should change. Bonds because it used to iterate over term, which is no longer allowed. Equity because it uses the max allowed term value.
I did not inlcude an explanation for why 5 is the max allowed term value in the vignette but would rather throw a message when terms are capped to 5 in the code because IMO this is so relevant that we should make sure the user really is aware of it.


data-raw/stress_test_arguments.R
```diff
@@ -9,7 +9,7 @@ stress_test_arguments <- tibble::tribble(
             "discount_rate",    "double",   "0.07",                     NA,    "-0.01",      "0.1",
   "div_netprofit_prop_coef",    "double",      "1",                     NA,      "0.8",        "1",
                "shock_year",    "double",   "2030",                     NA,     "2025",     "2035",
-                     "term",    "double",      "2",                     NA,        "1",       "10",
```
jacobvjk, line 12: My understanding was that we allow an input of max 10 years and that the call of the Merton function automatically puts any term > 5 into the 5 year bucket. I think this should already be happening. We should probably still notify users in that case

tests/testthat/_snaps/stress_test_arguments.md
```diff
@@ -13,6 +13,6 @@
       5 discount_rate           double    0.07    <NA>                 -0.01 0.1  
       6 div_netprofit_prop_coef double    1       <NA>                 0.8   1    
       7 shock_year              double    2030    <NA>                 2025  2035 
-      8 term                    double    2       <NA>                 1     10
```
jacobvjk, line 16: see above

jacobvjk (commented):
see comments

jacobvjk:
> This is a prep PR for allowing input of individual loans. I open it as a separate PR beforehand so it is possible to keep an overview of how it impacts results. Here snapshot tests for bonds and equity should change. Bonds because it used to iterate over term, which is no longer allowed. Equity because it uses the max allowed term value. I did not inlcude an explanation for why 5 is the max allowed term value in the vignette but would rather throw a message when terms are capped to 5 in the code because IMO this is so relevant that we should make sure the user really is aware of it.

agreed on throwing a message for the user

data-raw/stress_test_arguments.R
MirijaHa: But why would we allow using values >10 if we cap them to 5 anyway?

jacobvjk: we assumed that institutes regularly have such terms in their loan books. In the end I don't really care and I think this is something we can revisit with some user feedback to make a good decision

jacobvjk: Or rather: if we decide we cap at years here already, then do you add a section that will turn any longer terms into 5 years when reading i the user input? I think it may be too much of a burden to ask them to manually change all numbers greater than 5 before running the software

MirijaHa: we do not cap year here already. This is the default values that are chosen.
The capping of values >5 (that can still happen as soon as setting terms on user level is activated) is independent of this and not affected by the PR.
However I would say the only reason one would not set the maximum fallback value for terms to 5 is if terms > 5 are used somewhere in an uncapped way that I am not aware of?

jacobvjk: They are not being used anywhere. It's fine for me to cap them as early as possible, but I would suggest to do so in a way that allows users to keep individual longer terms in the input data and rather bucket them in the code plus returning a message: "200 loans in the loan book had input term > 5 years. These were automatically set to 5 years by the credit risk model."

MirijaHa:
closing becausse redundant

PR #281, opened 2021-12-21
Read and wrangle term

MirijaHa:


MirijaHa:
This PR

- prevents iterating over term
- sets max value for term to 5
- allows reading in company_term data. compare ST_EQUITY/inputs and ST_LOANS/input where i prepared respective input files (equity: all terms NA or 2, loans: all terms availble and in 1:10)

jacobvjk (approved):
lgtm

PR #282, opened 2021-12-22
Overlap checks

MirijaHa:


R/process_data.R
```diff
@@ -35,10 +35,21 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
     dplyr::filter(.data$scenario_geography %in% .env$scenario_geography_filter) %>%
     dplyr::filter(.data$ald_sector %in% .env$sectors) %>%
     dplyr::filter(.data$technology %in% .env$technologies) %>%
-    dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$end_year)) %>%
+    dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$start_year + .env$time_horizon)) %>%
```
MirijaHa, line 38: not sure if this is correct...

MirijaHa:
This PR shall give us more confidence in our input data.
It checks:

- that required levels on several variables are available
- that combinations of levels is exhaustive
- that in case of correlated variables (technology vs ald_sector) the mapping is as expected

This allowed me to also remove and now unneeded check for overlap of scenarios.

@fariiaakh this might be interestign for u too.

R/process_data.R
jacobvjk: seems more precise to me

tests/testthat/test-check_data.R
```diff
@@ -5,3 +5,56 @@ test_that("informative error is thrown if technologies are missing", {
...
+
+  test_data_checked <- check_sector_tech_mapping(
+    data = test_data,
+    sector_col = "sector"
+  )
+
+  expect_equal(test_data, test_data_checked)
+})
+
+test_that("error is thrown in case technology expected for ald_sector is missing.", {
```
jacobvjk, line 40: If a portfolio is only invested in a company that extracts Oil but not Gas, would this fail?

jacobvjk (commented):
see comments

tests/testthat/test-check_data.R
MirijaHa: nope, the check is only appleid to scenario and price data, so should not care about the portfolio

jacobvjk: fantastic

jacobvjk (approved):
lgtm

PR #283, opened 2021-12-22
reduce noise

cjyetman:
In my never-ending quest to reduce the noise in the output to the console so that it's easier to find/notice messages that actually matter, I've removed a few `print()` commands and added `quiet = TRUE` to `devtools::load_all()`

### DIFF
--- a/web_tool_external_stress_test.R
+++ b/web_tool_external_stress_test.R
@@ -23,7 +23,7 @@ source(file.path(stress_test_path, "R", "set_paths.R"))
 source("0_web_functions.R") # This script is sourced from PACTA_analysis, so path is correct
 source(file.path(stress_test_path, "R", "stress_test_model_functions.R"))
 
-devtools::load_all()
+devtools::load_all(quiet = TRUE)
 
 ################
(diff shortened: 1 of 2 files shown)


jacobvjk (commented):
Is there a log file we could write the messages to?

cjyetman:
you could use `write_log()` (with the assumption that [0_global_functions.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/194765169d3dddeecb13a73fa05cf35c399bc89d/0_global_functions.R) has been loaded and will therefore be available)

https://github.com/2DegreesInvesting/PACTA_analysis/blob/194765169d3dddeecb13a73fa05cf35c399bc89d/0_global_functions.R#L385-L396

cjyetman:
ok, I changed the `print()` statements to `write_log()` statements, except for the two really unnecessary ones (`print("Calculate Stress Test for Equity Portfolio")` and `print("Calculate Stress Test for Bonds Portfolio")`)

jacobvjk (approved):
lgtm

PR #284, opened 2021-12-23
first set of test cases

jacobvjk:
closes ADO 3322: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3322
**work in progress**

- I am adding a script to data-raw that will generate fake sample data sets including made up companies/portfolios as well as fake project agnostic data.
- exact directory and format of target files to be discussed

**companies/p4i style dataset**
- added a function that generates portfolio/companies in p4i format based on stylized cases of interest
- cases: that have one increasing and one decreasing technology each, differing in terms of market size, alignment, phase out policy, zero start value; companies with increasing or decreasing technologies only with different alignments
- it will have to be tested, if further diversification is needed to reflect companies without enough time to compensate
- possibly we need cases of companies that are strongly misaligned, slightly misaligned, exactly aligned, slightly overachieving in alignment, strongly overachieving in alignment. This is not yet taken care of
- scen_tech_prod (in fact all scenario related production variables) are now based on the scenario targets of the toy scenario data set for consistency

**exposures/p4i style dataset**
- added function that creates toy exposure data set in p4i style based on the companies/portfolio toy data

**scenarios**
- added function that generates toy scenario data for two scenarios and two sectors / eight technologies
- these are selected to match the companies in the sample portfolio
- at this stage, only scenario data in the format similar to the one in p4i is provided, so that the sample portfolio will not work for loans at this point, if the loan book has not been prepared yet

**cap factors**
- added function that generates toy scenario data for two scenarios and six technologies in the power sector

**prices**
- added functions that generate toy price data in wide (current) and long (future) formats for all relevant technologies and scenarios

**financial data**
- added function that generates toy financial data for each of the test companies
- The financial data are identical for all companies, as the test cases currently do not check the mechanism of the credit risk model, but rather the compensation mechanism

**exclusion**
- added function that creates an empty company exclusion data set in the format that would normally be used
- Company exclusion is not a test case, hence the empty data set.


jacobvjk:
In principle, the data sets generated here will run through as an equity project.
Some further fine tuning is needed.
Specifically, the way the data sets interact needs to be reviewed, so that the test companies behave as described in the task.
Some more consistency checks would be a nice to have

jacobvjk:
@MirijaHa @maurolepore I already tagged you for review here although this is still in draft mode. Some final decisions have to be made.
I am not sure I already want to store files generated by these test scripts before we check internally if there are any major issues with the compensation mechanism that should be fixed first.
But in principle, I think it could be nice to use these data sets as minimum user facing examples and once we are sure all edge cases of the compensation mechanism are covered, they might be used for integration tests.

jacobvjk:
also sorry for making this a pretty huge PR. I found it hard to wrap my head around how the scripts should be structured without seeing the interactions, so I put everything in one PR. Please let me know, if you want me to guide you through this.

jacobvjk:
this PR is move to the data repo https://github.com/2DegreesInvesting/r2dii.stress.test.data/pull/55

PR #285, opened 2021-12-27
reduce more noise

cjyetman:
following the same reasoning and strategy in #283, convert the last two `print()` statements to `write_log()` statements


jacobvjk (approved):
lgtm

PR #289, opened 2022-01-03
3131 Refactor less

maurolepore:
- Extract iteration_sequence()
- Refactor
- Refactor


R/run_stress_test.R
```diff
@@ -92,10 +92,7 @@ run_stress_test <- function(asset_type,
     iter_var = iter_var
   )
 
-  args_tibble <- tibble::as_tibble(args_list) %>%
-    dplyr::mutate(iter_var = .env$iter_var)
-
-  st_results_list <- purrr::map(1:nrow(args_tibble), run_stress_test_iteration, args_tibble = args_tibble)
+  st_results_list <- run_stress_test_iteration(args_list)
```
maurolepore, line 95: All friends of `run_stress_test()` now input the same single argument `args_list`. That list abstracts the details of which arguments it holds, and make it easy to add, remove, and change the arguments it contains.

R/run_stress_test.R
```diff
@@ -127,53 +124,35 @@ run_stress_test <- function(asset_type,
   cat("-- Exported results to designated output path. \n")
 }
 
-#' Iterate over stress test runs
-#'
-#' @param n Numeric.
-#' @param args_tibble  A tibble holding a set of params for
-#'   `run_stress_test_imp` per row.
-#'
-#' @return List of stress test results.
```
maurolepore, line 136: Internal functions don't need documentation and in this case its most maintainable to not do so.

MirijaHa (commented):
Already much nicer, thanks.
Before approving I have a question on the zzz file

- why did you add it? from the git history I think you did it to fix a problem with checks. I could not repro a problem in the checks, which problem did you encounter?
- am I right in thinking that calling globalVariables() here only has effect on the checks and cannot possible interfere with the actual code runs? 
- since I clearly do not understand the line I would argue this is a place where on oneliner on the WHY of the code would be rather beneficial?

maurolepore:
Thanks @MirijaHa for noting that. I now defined the global variables right next to where we need them. I think this is simpler and easier to maintain. 

> why did you add it? from the git history I think you did it to fix a problem with checks. I could not repro a problem in the checks, which problem did you encounter?

`list2env()` defines objects in the environment of the function that calls it. That way of defining objects is unconventional so R CMD check NOTEs: "Undefined global ... variables". `globalVariables()` solves the problem but my experiments and lil reading suggest you can't call it liberally:

- You can't call it inside the function that calls `list2env()` because by the time `globalVariables()` is called the package namespace is already compiled and locked.
- You can't call it outside the function and before `run_stress_test()` is defined because at that point the arguments to `run_stress_test()` are unknown.

So you must to call `globalVariables()` outside the function (a) and after we define `run_stress_test()` (b). It looks like R files are sourced alphabetically so R/zzz.R ensures (b). But then it's too far from where we use the variables and it creates confusion.

And so we now have these key elements in this rigid specific order:

```r
# R/run_stress_test.R
run_stress_test <- function(...

globalVariables(names(formals(run_stress_test)) ...
run_stress_test_imple <- function(args_list) {
  list2env(args_list, ...
```

> am I right in thinking that calling `globalVariables()` here only has effect on the checks and cannot possible interfere with the actual code runs?

Correct. It's like using `.data$ in `dplyr::filter(mtcars, .data$cyl < 4)` -- just to keep R CMD check happy.

> since I clearly do not understand the line I would argue this is a place where on oneliner on the WHY of the code would be rather beneficial?

Please see if the new location and one-line comment are clear and let me know.

R/run_stress_test.R
```diff
@@ -127,53 +124,37 @@ run_stress_test <- function(asset_type,
...
+    purrr::map(~dplyr::slice(args_tibble, .x)) %>%
+    purrr::map(run_stress_test_iteration_once)
+
+  return(out)
+}
+
+# Avoid R CMD check NOTE: "Undefined global functions or variables"
+globalVariables(c(names(formals(run_stress_test)), "iter_var"))
+run_stress_test_impl <- function(args_list) {
+  list2env(args_list, envir = rlang::current_env())
```
maurolepore, line 157: Here is the trick.

MirijaHa (approved):
lgtm

PR #291, opened 2022-01-04
3182 term processing

MirijaHa:
Since we rely on company to corporate bond ticker being correct i added an explicit check.
I had to make it conditional on asset type, because there is quite an inconsistency in pacta_results from the different asset_types.
BONDS: col id = corporate bond ticker
EQUITY col if = company id
LOANS col id = IDK, appears to be NA
🤷


MirijaHa:
git chaos, will reopen from new branch

PR #292, opened 2022-01-04
3465 ticker

MirijaHa:
Since we rely on company to corporate bond ticker being correct i added an explicit check.
I had to make it conditional on asset type, because there is quite an inconsistency in pacta_results from the different asset_types.
BONDS: col id = corporate bond ticker
EQUITY col if = company id
LOANS col id = IDK, appears to be NA


jacobvjk (approved):
lgtm
and yes, there is indeed no id column for loans at this point. p4b are thinking about adding the company_id there as well in the future

MirijaHa:
> lgtm and yes, there is indeed no id column for loans at this point. p4b are thinking about adding the company_id there as well in the future

any change they could switch to making id a semantically clear name? I.e. turning it to company_id or corporate_bond_ticker, depending on what it holds?

jacobvjk:
> > lgtm and yes, there is indeed no id column for loans at this point. p4b are thinking about adding the company_id there as well in the future
> 
> any change they could switch to making id a semantically clear name? I.e. turning it to company_id or corporate_bond_ticker, depending on what it holds?

well that's from P4I which usually makes everything a lot more complicated..

MirijaHa:
> > > lgtm and yes, there is indeed no id column for loans at this point. p4b are thinking about adding the company_id there as well in the future
> > 
> > 
> > any change they could switch to making id a semantically clear name? I.e. turning it to company_id or corporate_bond_ticker, depending on what it holds?
> 
> well that's from P4I which usually makes everything a lot more complicated..

I also assume from P4I logic it somewhat makes sense to have a column called id, which holds different information depending on asset type, as from bonds the ticker and for equity the company is the relevant leve..

PR #293, opened 2022-01-04
3182 activate term use

MirijaHa:
Coming to the end with enabling user provided terms.

This PR actually uses them.

If no terms are provided fallback term is used, same applies to missings.

Please note the following expectations on snapshot tests:

- bonds: no fail, I did not switch on use_company_terms here
- loans: failure: all terms are provided, sampled from 1:10, so results will change
- equity: no fail: all terms are either 5 (current input arg value) or NA (will use fallback 5)


jacobvjk (approved):
a-ma-zing

PR #294, opened 2022-01-05
3182 explict capping

MirijaHa:
Second last PR on the term fun (will add sth to vignette on how to prepare the term data).
This PR make sure that all processing that happens with the company_terms input happens in one place and is explictly accompanied with a message to the user.

Dear @jacobvjk  please check above all if the messages are correct.

Expectation: snapshot tests should pass


jacobvjk (approved):
nice one, totally lgtm
but would it be possible/reasonable, to use the new cap function here as well?
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/e283dbc97ca164615277f76c2c3f6130da93d570/R/company_expected_loss.R#L65

jacobvjk:
> > nice one, totally lgtm but would it be possible/reasonable, to use the new cap function here as well?
> > https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/e283dbc97ca164615277f76c2c3f6130da93d570/R/company_expected_loss.R#L65
> 
> mhhh, i deleted the row u mentioned from the PR, since there is no way that at that stage there is still terms > 5. If they come from company_term they are capped in data processing, and the maximum for fallback cap is 5 as well. However someone could of course not be awar of this and adjust the allowed maximum for the fallback. Maybe I add an explicit comment to stress test argumnts why the upper cap of 5 needs to be capped?

You can, but at the same time I think it is obvious that if you change things beyond a given min/max, unexpected things may happen.. no?

jacobvjk (approved):
makes sense

PR #297, opened 2022-01-07
Vignette fin data

MirijaHa:
Closes: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3529
This PR adds a vignette for using custom financial data. 
Note: this is unnecessarily clumsy due to:

- inconsistencies in PACTA results on company_id
- using of corporate bond ticker for merges for bonds.
I would shortly review if corporate bond ticker can be omitted from financial data and the simplify the workflow if possibly, however like this we  at least provide a functioning option.


jacobvjk (approved):
lgtm

PR #302, opened 2022-01-11
`st_read/process()` now handle `capacity_factor_power`

maurolepore:
Relates to #3116
Extends #300

Handle capacity_factor_power﻿

### DIFF
--- a/R/run_stress_test.R
+++ b/R/run_stress_test.R
@@ -171,7 +171,7 @@ read_and_process <- function(args_list) {
   cat("-- Reading input data from designated input path. \n")
   data <- append(
     st_read_specific(input_path_project_specific, asset_type),
-    st_read_agnostic(input_path_project_agnostic, asset_type)
+    st_read_agnostic(input_path_project_agnostic)
   )
 
   start_year <- get_start_year(data)
@@ -345,7 +345,7 @@ st_read_specific <- function(dir, asset_type) {
   return(out)
 }
 
-st_read_agnostic <- function(dir, asset_type) {
+st_read_agnostic <- function(dir) {
   out <- list(
     capacity_factors = read_capacity_factors(capacity_factor_file(dir))
   )
--- a/R/run_stress_test.R
+++ b/R/run_stress_test.R
@@ -169,21 +169,13 @@ read_and_process <- function(args_list) {
   )
 
   cat("-- Reading input data from designated input path. \n")
-  data <- st_read(input_path_project_specific, asset_type)
+  data <- append(
+    st_read_specific(input_path_project_specific, asset_type),
+    st_read_agnostic(input_path_project_agnostic, asset_type)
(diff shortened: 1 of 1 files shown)


R/run_stress_test.R
```diff
@@ -371,9 +374,20 @@ st_process <- function(data, asset_type) {
...
     process_sector_exposures(asset_type = asset_type)
 
+  capacity_factors_power <- data$capacity_factors %>%
+    process_capacity_factors_power(
+      scenarios_filter = scenarios_filter,
+      scenario_geography_filter = scenario_geography_filter_lookup,
+      technologies = technologies_lookup,
+      start_year = start_year,
+      end_year = end_year_lookup
+    )
```
maurolepore, line 384: I see we go from `capacity_factor` to `capacity_factor_power`. Is this correct? I assume we read data from every sector, then process only data in the power sector. If that's the case, I would prefer to expose the step that picks the power sector specifically. Does it make sense?

```r
data$capacity_factors %>%
  pick_power() %>%
  process_capacity_factor_power()
```

MirijaHa: Apparently capacity_factors (currently) only holds capacity_factors for the power sector ATM. This might be extended but I have no background knowledge on how likely and how soon that would happen, maybe @jacobvjk  knows more.
While the status quo is as it is I would suggest renaming read_capacity_factors to read_capacity_factors_power. If doing so please make sure it does not interfere with the webtool though/adjust webtool accordingly.

R/run_stress_test.R
```diff
@@ -391,6 +405,11 @@ sector_exposures_file <- function(dir) {
   return(out)
 }
 
+capacity_factor_file <- function(dir) {
```
MirijaHa, line 408: I would suggest to at some point move all filename creation, reading wrappers and processing wrappers out of the run_ file to some more specialised files

R/run_stress_test.R
```diff
@@ -353,14 +345,25 @@ st_read <- function(dir, asset_type) {
   return(out)
 }
 
+st_read_agnostic <- function(dir, asset_type) {
```
MirijaHa, line 348: asset_type is not a necceasary input arg to asset_read_agnositc (at least as of now)

MirijaHa (commented):
small comments

R/run_stress_test.R
maurolepore: I defer it to https://github.com/2DegreesInvesting/r2dii.climate.stress.test/issues/307.

R/run_stress_test.R
maurolepore: Defer it to #308

R/run_stress_test.R
maurolepore: Thanks for noting that! Done.

PR #304, opened 2022-01-11
`st_read/process()` now handle `excluded_companies`

maurolepore:
Relates to ab#3116
Extends and waiting for #302﻿

### DIFF
--- a/R/run_stress_test.R
+++ b/R/run_stress_test.R
@@ -177,16 +177,6 @@ read_and_process <- function(args_list) {
   start_year <- get_start_year(data)
   scenarios_filter <- scenarios_filter()
 
-  df_price <- read_price_data_old2(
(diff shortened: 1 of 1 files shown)


R/run_stress_test.R
```diff
@@ -215,7 +210,7 @@ read_and_process <- function(args_list) {
     process_company_terms(fallback_term = term)
 
   cat("-- Processing input data. \n")
-  processed <- st_process(data, asset_type)
+  processed <- st_process(data, asset_type, company_exclusion)
```
MirijaHa, line 213: number of arguments is starting to grow, how about using names in the call to st_process?

MirijaHa (approved):
1 small suggestion

R/run_stress_test.R
maurolepore: Done

PR #310, opened 2022-01-12
use newest pckdown workflow

MirijaHa:
This uses the newest pkgdown workflow from V2, please note that changes in comparison to master are a bit bigger on this particular one.


cjyetman (approved):
don't really know much about these pkgdown actions, but generally looks good to me

maurolepore (approved):
LGTM

MirijaHa:
This PR keeps producing errors with R installtion on various OS and versions. I would argue the gain is not worth finding out why so will close the PR

PR #312, opened 2022-01-13
update web tool script CI

cjyetman:
Some big changes are coming to PACTA, and in prep for that I added a dependency on `{vroom}`. This is a stopgap fix to allow the Action to run without error, because a completely new GH Action to test the web tool scripts is on its way, which will based on a completely new container, which will be aligned with a completely new Dockerfile which will package everything for the website.


maurolepore (approved):
LGTM

PR #313, opened 2022-01-13
Fix tmsr smsp

jacobvjk:
closes ado 3384 (https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3384) and 3109 (https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3109)

**This PR:**
- implements the use of SMSP for increasing technologies when extending scenario trajectories
- implementation of SMSP means that 0 start values are no longer an issue and hence apply_filters is removed from the user workflow
- ensures that companies phasing out technologies within the forecast period are not forced to reintroduce them due to the compensation mechanism and that production targets cannot go below 0.
- uses the tech mix in the last year of the production forecast to derive company-technology exposures. This is closer to the year of shock and it allows for companies that start building out within the forecast period but after the start year to retain that technology.
- adjusts the web tool to use old functions and not interfere with results

**Note:**
- at this point, inner-joining the portfolio exposures and the company results will still kick out company_technology combinations with 0 production at the end of the forecast period. This is because we use plan_carsten from within that period to get the exposure. If that exposure is 0, the holding is not retained
- For the time being that means that no company can enter a market that has not built out at least some capacity at the end of the production timeframe
- If this needs to change, we would have to rethink how we define technology exposure at the company level


jacobvjk:
> > also how come we still have a macos check running here?
> 
> this is a public repo so no charges apply

ah true...

jacobvjk:
The note turns out to be a methodological question about which year we should use to approximate company/technology level exposure. Discussing this and adapting the PR accordingly, once we have a solution

jacobvjk:
btw this will definitely change the results of the snapshot. I am still not entirely comfortable using the snapshot tests due the initial difficulties on my machine and a lack of relevant PRs from my side since then.

R/apply_filters.R
```diff
@@ -22,7 +22,7 @@
 #' @param start_analysis Numeric. A vector of length 1 indicating the start
 #'   year of the analysis.
 
-apply_filters <- function(data,
+apply_filters_old <- function(data,
                           investor = NULL,
                           sectors = NULL,
                           technologies = NULL,
```
maurolepore, line 28: Style please.

maurolepore: Does `_old` mean "dead code"? If so, can we delete it?

You can always find an old function, say `apply_filters`, with something like this:

```r
git log -S "apply_filters <- function" -p
```

R/apply_filters.R
```diff
@@ -87,3 +87,98 @@ apply_filters <- function(data,
...
+      key = paste(.data$company_name, .data$id, .data$ald_sector, .data$technology, .data$scenario_geography)
+    ) %>%
+    # ADO 2393 - running distinct extracts the key as a variable with unique
+    # values to be used for removal of invalid entries below
+    dplyr::distinct(.data$key)
+
+  data <- data %>%
+    dplyr::mutate(
+      key = paste(.data$company_name, .data$id, .data$ald_sector, .data$technology, .data$scenario_geography)
+    ) %>%
```
maurolepore, line 181: This overflows the 80 charactes (I think) and appears twice. Maybe write a small helper `paste_xyz_cols(data)` -- it  would remove the duplication and fit in a line.

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +109,153 @@ extend_scenario_trajectory <- function(data,
...
+#'   parameters of the analysis, including the business as usual and target
+#'   scenarios, the relevant scenario geography and time frame for each of the
+#'   technologies.
+#' @param start_analysis Numeric. A vector of length 1 indicating the start
+#'   year of the analysis.
+#' @param end_analysis Numeric. A vector of length 1 indicating the end
+#'   year of the analysis.
+#' @param time_frame Numeric. A vector of length 1 indicating the number of years
+#'   for which forward looking production data is considered.
+extend_scenario_trajectory <- function(data,
```
maurolepore, line 135: If this function is not to be exported then add `@noRd` to stop roxygen2 from generating the documentation. It usually helps keep R CMD check happy.

```r
#' @noRd
```

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +109,153 @@ extend_scenario_trajectory <- function(data,
...
+      .data$scenario_geography, .data$source, .data$units
+    ) %>%
+    dplyr::mutate(initial_sector_production = dplyr::first(.data$plan_sec_prod)) %>%
+    dplyr::ungroup()
+
+  data <- data %>%
+    dplyr::mutate(
+      scen_tech_prod = dplyr::if_else(
+        .data$direction == "declining",
+        .data$initial_technology_production * (1 + .data$fair_share_perc), # tmsr
```
maurolepore, line 236: Maybe convert the comment in a lil helper?

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +109,153 @@ extend_scenario_trajectory <- function(data,
...
+        .data$initial_technology_production + (.data$initial_sector_production * .data$fair_share_perc) # smsp
```
maurolepore, line 237: Maybe convert the "smsp" comment in a lil helper.

web_tool_stress_test.R
```diff
@@ -324,7 +324,7 @@ if (file.exists(file.path(results_path, pf_name, paste0("Equity_results_", calcu
       # Convert capacity (MW)to generation (MWh) for power sector
       equity_annual_profits <- pacta_equity_results %>%
         convert_cap_to_generation(capacity_factors_power = capacity_factors_power) %>%
-        extend_scenario_trajectory(
+        extend_scenario_trajectory_old(
```
maurolepore, line 327: Similarly to a comment above, reflect on whether or not we need old code.

maurolepore (approved):
LGTM. Because I'm detached from the concepts all my comments are nit -- feel free to disregard.

R/apply_filters.R
jacobvjk: _old means it is used by the web script in the interactive report, because we do not want to interfere with results of past projects.

web_tool_stress_test.R
jacobvjk: it's still in use

R/apply_filters.R
jacobvjk: done

R/apply_filters.R
maurolepore: Thanks, so it's definitely not dead code. Then for maximum backward compatibility it seems easier to use `_new()` for new stuff rather than `_old` for old stuff -- that way you don't need to change any existing code.

MirijaHa: quick general comment: The plan is to delete old webtool functions (i.e. _old) functions as soon as we are allowed to. That is why we prefer to give those ones the clumsy name _old then to use _new on the functions that will soon be the only ones.

MirijaHa:
will review when merge conflicts are resolved

MirijaHa (changes requested):
Calculations and snapshot tests fail 😿 . I would suggest first checking if it was induced by mistakes in the last merge commit.

R/calculate.R
```diff
@@ -108,7 +111,7 @@ calculate_exposure_by_technology_and_company <- function(asset_type,
 
   exposure_by_technology_and_company <- input_data_list$pacta_results %>%
     dplyr::filter(
-      .data$year == .env$start_year,
+      .data$year == .env$start_year + .env$time_horizon,
```
MirijaHa, line 114: I take it this part was sneaked into the PR based on a discussion on slack? I would like to kindly encourage you to:
- revert affected change in this PR
- open a new PR inlcuding only that change
- running a change check or snapshot test on it to quantify the impact

Background: This particular change is small and easy to review, so I could elt it go, but it will affect all results I guess. This makes it impossible to test expectation on changes of the tmsr smsp fix.

MirijaHa (changes requested):
Pls remove not PR related commit

R/apply_filters.R
```diff
@@ -71,19 +158,34 @@ apply_filters <- function(data,
   remove <- data %>%
     dplyr::filter(
       .data$plan_tech_prod == 0,
-      .data$year == .env$start_analysis
-    ) %>%
-    dplyr::mutate(
-      key = paste(.data$company_name, .data$id, .data$ald_sector, .data$technology, .data$scenario_geography)
+      .data$year == .env$start_analysis,
+      .data$technology %in% .env$declining_technologies
```
MirijaHa, line 162: Just to go sure, the difference between apply_filters and apply_filters old is that in apply filters all decling technologies are removed?

MirijaHa: if yes maybe also add this to the description to disambiguiate it from apply_fiters_olds

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +110,154 @@ extend_scenario_trajectory <- function(data,
...
+
+
+#' Extend the scenario pathways based on the fair share approach (now known as
+#' market share approach).  We first ensure that all scenarios are extended to
+#' the same point in time (start of the analysis plus forecast time frame.
+#' Then we take the first value of the production on company level as the start
+#' value and from the second year onward, we calculate the level of production
+#' for a company under the scenarios at hand by applying the relative changes
+#' for the respective years to the absolute values of the previous year.
+#'
```
MirijaHa, line 122: please make more specifc here how it extends extend_scenario_trajectory

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +110,154 @@ extend_scenario_trajectory <- function(data,
...
+#'   technologies.
+#' @param start_analysis Numeric. A vector of length 1 indicating the start
+#'   year of the analysis.
+#' @param end_analysis Numeric. A vector of length 1 indicating the end
+#'   year of the analysis.
+#' @param time_frame Numeric. A vector of length 1 indicating the number of years
+#'   for which forward looking production data is considered.
+#' @noRd
+extend_scenario_trajectory <- function(data,
+                                       scenario_data = NULL,
```
MirijaHa, line 138: I am conceptually to detached to understand/review this, sorry. If you are not comfortable merging this without approval let us synch on alternative strategy to increase confidence in the solution by tessting/running thorugh solution together

R/run_stress_test.R
```diff
@@ -331,9 +336,12 @@ st_read_agnostic <- function(dir, start_year) {
   return(out)
 }
 
-st_process <- function(data, asset_type, company_exclusion) {
+st_process <- function(data, asset_type, company_exclusion, scenario_data) {
```
MirijaHa, line 339: scneario_data is a list entry in data here

R/run_stress_test.R
```diff
@@ -331,9 +336,12 @@ st_read_agnostic <- function(dir, start_year) {
...
   start_year <- get_start_year(data)
   scenarios_filter <- scenarios_filter()
+  declining_technologies <- scenario_data %>%
```
MirijaHa, line 342: edgecase question, what happens if there are not decllining technologies, does everthing run smooth?

MirijaHa (commented):
Some more comments.
As stated I not not have sufficient understand to review the content. 

What could be helpful though:
- **interactive test**: For which data do you expect changes with this release? More importantly, is there data for which you expect no changes?
- **unit tests**: can u identify minimal cases, e.g. from the toy data  and formulate expectations on their behavior in the form of unit tests?

R/calculate.R
jacobvjk: done https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/316

R/run_stress_test.R
jacobvjk: updated

R/extend_scenario_trajectory.R
jacobvjk: overhauled

R/run_stress_test.R
jacobvjk: In the portfolio or in the scenario data? There will not be a scenario data set without declining technologies. Will check on the portfolio level though

R/extend_scenario_trajectory.R
jacobvjk: done

jacobvjk:
regarding the overall test:

- on the company_tech level, there is no relative change for declining technologies, there almost always is a change for increasing technologies. This makes sense, as it simply reflects the tmsr/smsp handling.
- For the companies that were formerly excluded, we get no differences in results regardless of whether they are explicitly excluded or not, since the phase out is now automatically detected.
- One company was apparently erroneously excluded, this company now remains in the results, when the exclusion option is set to false, as the automatic detection correctly does not include that combination.

R/run_stress_test.R
jacobvjk: Same impact on company, if only increasing technologies are given

R/apply_filters.R
jacobvjk: it is about identifying which company/technology combination has zero production value in the start year and removing that. Given the recent changes, this is not necessary anymore and should be removed altogether

R/wrangle_and_check.R
```diff
@@ -80,16 +80,7 @@ wrangle_and_check_pacta_results <- function(pacta_results, start_year, time_hori
...
-    apply_filters(
-      investor = investor_name_placeholder,
-      sectors = sectors_lookup,
-      technologies = technologies_lookup,
-      scenario_geography_filter = scenario_geography_filter,
-      scenarios = scenarios_filter,
-      allocation_method = allocation_method,
-      start_analysis = start_year
-    )
+    dplyr::mutate(plan_tech_prod = dplyr::if_else(is.na(.data$plan_tech_prod), 0, .data$plan_tech_prod))
```
jacobvjk, line 83: I surfaced this mutate from the previous apply_filters, all actual filters were moved to the processing function (they were mostly redundant)

jacobvjk:
simplified the pacta data processing, apply filters not necessary anymore in user workflow. I checked along the way that results do not change based on removing the function call

R/apply_filters.R
jacobvjk: apply filters was removed

R/run_stress_test.R
MirijaHa: I do not understand the last remark, so just to go sure. Let us assume we have portfolio that holds only 1 company in the renewables sector. The analysis will still run?

R/wrangle_and_check.R
MirijaHa: I compared pacta_results between master and PR version for type bond and see that PR version has quite a bunch of more rows (master: 7836, PR: 9552).

R/run_stress_test.R
jacobvjk: The test was: portfolio with one company in the power sector. This company has assets in the Nuclear and Renewables technologies, both of which are `"increasing"`. Hence no, declining techs given. Still runs. But: the newest version should not contain the `declining_technologies` variable anymore anyway

R/wrangle_and_check.R
MirijaHa: compare pacta_results_new %>% dplyr::anti_join(pacta_results_old) %>% View(), I can not see a clear pattern, so is this expected?

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +108,177 @@ extend_scenario_trajectory <- function(data,
...
+      .data$scen_tech_prod
+    ) %>%
+    dplyr::filter(.data$year <= .env$start_analysis + .env$time_frame) %>%
+    dplyr::group_by(
+      .data$investor_name, .data$portfolio_name, .data$id, .data$company_name,
+      .data$ald_sector, .data$technology, .data$scenario, .data$allocation,
+      .data$scenario_geography
+    ) %>%
+    dplyr::mutate(
+      initial_technology_production = dplyr::first(.data$plan_tech_prod),
```
MirijaHa, line 181: maybe do some explicit sorting earlier, because I thinkyou depend on sorting being chronological

R/extend_scenario_trajectory.R
```diff
@@ -109,3 +108,177 @@ extend_scenario_trajectory <- function(data,
...
+
+  validate_data_has_expected_cols(
+    data = scenario_data,
+    expected_columns = c(
+      "source", "technology", "scenario_geography", "ald_sector", "units",
+      "scenario", "year", "direction", "fair_share_perc"
+    )
+  )
+
+  data <- data %>%
```
MirijaHa, line 167: it would really drastically imcrease the readibility of this function if you moved each block where at the end of a pipe u assign to data (and some calculation, not pure wrangling, happens) into its own function. As they are interna it would be ok to not document them (a good name is already helpful enough) so this should be doable very quickly.

MirijaHa (commented):
some comments

R/extend_scenario_trajectory.R
jacobvjk: done

R/extend_scenario_trajectory.R
jacobvjk: done

jacobvjk: will check the diffs tomorrow. The resulting files are rather similar in size though

R/wrangle_and_check.R
MirijaHa: ok

R/extend_scenario_trajectory.R
MirijaHa: beautiful

MirijaHa (approved):
🕺

PR #316, opened 2022-01-28
use exposure from last year of prod forecast

jacobvjk:
preparation for tmsr/smsp fix

This PR:
- uses the last year of the production forecast to derive company/technology level exposures.
- any company for which the production share within their ald_sector changes over that time frame will have a different overall loss. However the percentage change on company_technology level should remain the same

### DIFF
--- a/R/calculate.R
+++ b/R/calculate.R
@@ -90,8 +90,11 @@ calculate_annual_profits <- function(asset_type, input_data_list, scenario_to_fo
 #'
 #' @return A tibble holding exposure_by_technology_and_company.
 calculate_exposure_by_technology_and_company <- function(asset_type,
-                                                         input_data_list, start_year,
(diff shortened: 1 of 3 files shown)


MirijaHa (commented):
Before i approve, did you verify that your expectations hold true?

jacobvjk:
additional expectation: companies that fade out production over the 5 year horizon will be kicked out

jacobvjk:
tested assumption:
- with toy data company 115 (all exposures stay stable) nothing changes, for other toy data companies with changes in production composition, the losses change.
- both companies (107 and 114) which fade out a technology (no matter if high or low carbon, have this technology removed)
- the relative magnitude of the shocks remains unchanged

MirijaHa (approved):
yolo

PR #318, opened 2022-02-02
Intermediate results

MirijaHa:
This closes https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3121

FYI @maurolepore , this add a new results  so snaps change
FYI @jacobvjk The new result type (compare e.g. last run of bonds results) contains more rows than other result types on company level, is that expected?

FYI @jakubcervenka notifying u as this introduced a new kind of results.


jacobvjk:
@MirijaHa yes, more rows are expected, as we provide year by year numbers in this case

PR #319, opened 2022-02-03
Fallback term

MirijaHa:
This closes remaining aspects from https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3835 (adding logline if company-terms are used, exposing fallback_term instead of term to avoid misleading information in results. The latter was slightly trickier than expected, since as soon as fallback_term is integrated in the model it has to be called term, so I could not simply search and replace all occurences.

Snapshot tests fail expectedly.


jacobvjk (approved):
lgtm

PR #320, opened 2022-02-03
Result namings

MirijaHa:
Renamed the output files as agreed.
Please check if you are fine with the namings, above all for market risk. After merging it I will also improve the namings downstream but am not doing that now to keep PR simple.


jacobvjk (approved):
lgtm

PR #321, opened 2022-02-03
Naming follow up

MirijaHa:
Follow up on results naming. Also used the naming convention in earlier steps for easier orientation


jacobvjk (commented):
pls merge master, then good to go

MirijaHa:
FYI @maurolepore  this merge will lead to failing snapshot tests because names of result files changes

PR #323, opened 2022-02-09
Prep scen geos

MirijaHa:
this is PR1 from https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/2400

After processing it is checked, if requested scenario_geographies (currently only global) are present in all affected data.

Review is not time critical.


jacobvjk (approved):
lgtm

PR #326, opened 2022-02-10
Revert overlap check

MirijaHa:
Another one for prepping code for flexible scenario geographies.
It turns out there is already a validation on dataset level, if all required levels of variables are given, this make the overlap checks introduced in https://github.com/2DegreesInvesting/r2dii.climate.stress.test/commit/29ad00de64c2b23561cdc77722724949e152eedf obsolete, instead I made the error message of the existing validation more helpful.

I also added a function to stop calculations if any needed input data hold 0 rows after filtering, as this is becoming more likely when allowing choice of geography.

Also I included 2 small fixes:

- fixed a typo leading to partial matching
- removed some unneeded function args


jacobvjk (approved):
lgtm
also looks like I didn't fully clean upo wrangle_and_check with the latest PR, so thanks for that

PR #328, opened 2022-02-10
Rm production shock perc

MirijaHa:
AS it does not affect webtool anymore removed NA variable from user flow results


jacobvjk (approved):
yay

PR #330, opened 2022-02-10
Adjust profits in late buildout

jacobvjk:
closes ADO 3530: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3530

**This PR:**
- calculates the ratio of how much of their target per technology a company has fulfilled by the end of the production forecast in order to...
- have a proxy for how much of a climate laggard the company is
- this ratio is then used to adjust the net profit margin of the ADDITIONAL production that climate laggards need to create in increasing technologies in order to compensate their carbon budgets.
- This is done to account for the fact that sudden upscaling of production does not come for free and lagging in increasing technologies is thus no incentive.
- Since these adjustments change the profit calculation, I renamed the old one to `calculate_net_profits_old()` and adapted the web script
- I put the new one into the file `R/financial_functions.R`, as I expect more functions in that area to change and think this grouping will make it easier to navigate

**NOTE:** initially we planned to also adjust the net profit margin on surplus production based on the relative market share of a company in that technology. Since the preparation of that data points to some open issues, we opted to move that adjustment to a follow up task: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/4109


R/extend_scenario_trajectory.R
```diff
@@ -132,14 +132,18 @@ extend_scenario_trajectory_old <- function(data,
...
+#'   scenario
+
 #' @noRd
 extend_scenario_trajectory <- function(data,
                                        scenario_data = NULL,
                                        start_analysis = NULL,
                                        end_analysis = NULL,
-                                       time_frame = NULL) {
+                                       time_frame = NULL,
+                                       baseline_scenario) {
```
MirijaHa, line 146: nit: it is recommendable to put args w/o defaults first.

R/extend_scenario_trajectory.R
```diff
@@ -372,3 +384,92 @@ handle_phase_out_and_negative_targets <- function(data) {
...
+#'   years for which forward looking production data is considered.
+#' @param baseline_scenario Character. A vector of length 1 indicating baseline
+#'   scenario
+#'
+#' @noRd
+calculate_proximity_to_target <- function(data,
+                                          start_analysis,
+                                          time_frame,
+                                          baseline_scenario) {
+  data <- data %>%
```
MirijaHa, line 409: this function throws warnings (no non-missing arguments to max; returning -Inf) at least for me locally, also I would argue it is unnecessarily complicated (unless I am overlooking a complexity).
How about:
- filtering data for the needed lines (year = .env$start_analysis + .env$time_frame, scenario = shock), 
- calculating proximity_to_target with some nice and tidy mutate statements
- right joining this to data

R/financial_functions.R
```diff
@@ -0,0 +1,28 @@
+#' Calculates annual net profits on the company-technology level for the
+#' baseline and late and sudden scenarios. Climate laggards which need to build
+#' out their production in increasing technologies to compensate for their
+#' missed targets, are "punished" by adjusting the net profit margin on their
+#' additional build out based on their proximity to target and market share within
+#' the given technology. This ensures that late build out will not
```
MirijaHa, line 6: I would argue it would be good to provided a more concrete information how the punishment happens.

R/financial_functions.R
```diff
@@ -0,0 +1,28 @@
...
+#'   under baseline and late and sudden, market prices/costs, company net profit
+#'   margins, the proximity to target in the production forecast period and an
+#'   indication of the direction of the technology.
+calculate_net_profits <- function(data) {
+  data <- data %>%
+    dplyr::mutate(
+      production_compensation = .data$late_sudden - .data$baseline,
+      net_profits_baseline = .data$baseline * .data$Baseline_price * .data$net_profit_margin,
+      net_profits_ls = dplyr::if_else(
+        .data$direction == "declining",
```
MirijaHa, line 21: just playing devils advocate, why is it not neccesary to also do some proximity to target-dependent handling for increasing technologies? Just asking because this looks assymetric.

MirijaHa (commented):
Some small remarks.
I also noticed some snapshot tests did show some big impact on results.
Did you do a systematic change check on the test porfolios and if yes are changes within an acceptable range?

R/financial_functions.R
jacobvjk: The contrary is happening. If the technology is `declining`, the basic profit function is used. if it is not `declining` it will get penalised based on the proximity.
This is not happening for decreasing technologies, because a compensation for decreasing technologies is a punishment on the profits per construction. It means less production for laggards, hence less profits.
For increasing, the inverse is the case. Compensation means additional production for laggards. What this punishment is doing is to say that this additional production cannot simply be passed to the same profit function because that would reflect unrealistic and perverse incentives. Not building out such technologies and then being forced to increase them would lead to a sudden need for strong investments and more costs though, which is what we are approximating here

jacobvjk:
@MirijaHa I did a systematic check on the toy data set. The summary of which you can find in the st_methodology channel and an email I just forwarded to you. My assessment is that these changes are reasonable and happen only where expected. But the final verdict on the acceptability of the impact is out for discussion in the team

jacobvjk:
methodology wise, we agreed to merge this for now

R/extend_scenario_trajectory.R
```diff
@@ -372,3 +379,91 @@ handle_phase_out_and_negative_targets <- function(data) {
...
+#'
+#' @noRd
+calculate_proximity_to_target <- function(data,
+                                          start_analysis,
+                                          time_frame,
+                                          baseline_scenario) {
+  production_changes <- data %>%
+    dplyr::filter(
+      dplyr::between(
+        .data$year, .env$start_analysis, .env$start_analysis + .env$time_frame
```
MirijaHa, line 407: I think if also filtering for scenario the calculations could still be further simplified and maybe even do without all the grouping. After all we are then at 1 row per group.
I know that this would become a problem if one allowed e.g. > 1 investor, but I largely doubt that that will ever happen, so in order to make this implementation easy one could also just add a quick check that each "group" holds only 1 line and then do without the groupings..

MirijaHa (changes requested):
1. I still get the warning ```no non-missing arguments to max; returning -Inf``` , if you do not we should find out why
2. what is returned from calculate_proximity_to_target changed between now and the last time I reviewed, which points to a potential problem

R/extend_scenario_trajectory.R
jacobvjk: simplified, warnings are gone now

R/calculate.R
```diff
@@ -38,7 +38,8 @@ calculate_annual_profits <- function(asset_type, input_data_list, scenario_to_fo
       scenario_data = input_data_list$scenario_data,
       start_analysis = start_year,
       end_analysis = end_year,
-      time_frame = time_horizon
+      time_frame = time_horizon,
+      target_scenario = shock_scenario_lookup
```
MirijaHa, line 42: could you explain again what the difference between target and shock scenario is? In the results for company_trajectories they are 2 different variables, practically they are the same I guess and here the naming is mixed>

MirijaHa (approved):
FYI, the data generated with calculate_proximity_to_target differs with the version from your original PR, but only on the column proximity_to_target (old: -inf, new: NaN) so I guess that is no problem

jacobvjk:
> FYI, the data generated with calculate_proximity_to_target differs with the version from your original PR, but only on the column proximity_to_target (old: -inf, new: NaN) so I guess that is no problem

that sounds unproblematic to me

PR #332, opened 2022-02-15
first overhaul of vignettes

jakubcervenka:


MirijaHa (commented):
Before reviewing this in detail some quick general remarks:
- In the ticket it was to my understanding only requested "Add a section to the manual that explains the fundamentals (of P4B) on a high level and links clearly to the relevant material."  A PR should never include more content than is specified in the ticket because otherwise reviewers cannot reasonably review. As a one time solution now I would  providing a clear PR comment so that the reviewers have a benchmark.
- Please look at the diff and see if all changes included are really supposed to be there. This PR reverts quite a few changes from last PRs, so I am not sure if that is intended or due to mistakes when solving merge conflicts
- the included readme covered the minimum of what is required for.a readme for an R package, so we really should not reduced it further, so please undo the deletions to the README.

jakubcervenka:
Ok I get why the readme needs to have information on how to install. When you look at the vignettes, the readme is the first thing a user sees. It does not make much sense to tell the user how to install before seeing the Prerequisites. We can take all the information from the Prerequisites vignettes and put it in the Readme but then we would make that a fair bit longer

jacobvjk:
the problem of the prerequisites will be solved, once we have the toy data published. So in principle, I would argue the structure of the readme makes sense even without the prerequisites in here. We could point out more prominently that at this time, one needs to request a prerequisites document for inputs though

jakubcervenka:
Closing as there are conflicts with earlier commits. I will open a new PR

PR #333, opened 2022-02-15
rename output columns

jacobvjk:
closes ADO 3126

This PR:
- renames the columns in the output files to more understandable terminology
- updates the vignettes
- remove hard coded scenario specific columns form trajectory output (SDS, NPS....)

### DIFF
--- a/R/wrangle_and_check.R
+++ b/R/wrangle_and_check.R
@@ -339,6 +339,26 @@ wrangle_results <- function(results_list, sensitivity_analysis_vars) {
       .data$analysed_sectors_value_change, .data$portfolio_aum,
       .data$portfolio_value_change_perc, .data$portfolio_value_change,
       .data$exclude, !!!rlang::syms(sensitivity_analysis_vars)
+    ) %>%
+    dplyr::rename(
+      company_technology_exposure = .data$tech_company_exposure,
+      company_technology_percent_value_change = .data$VaR_tech_company,
+      company_technology_absolute_value_change = .data$tech_company_value_change,
+      company_exposure = .data$company_exposure,
+      company_percent_value_change = .data$VaR_company,
+      company_absolute_value_change = .data$company_value_change,
+      technology_exposure = .data$technology_exposure,
+      technology_percent_value_change = .data$VaR_technology,
+      technology_absolute_value_change = .data$technology_value_change,
+      sector_exposure = .data$sector_exposure,
+      sector_percent_value_change = .data$VaR_sector,
+      sector_absolute_value_change = .data$sector_value_change,
+      analysed_portfolio_exposure = .data$analysed_sectors_exposure,
+      analysed_portfolio_percent_value_change = .data$VaR_analysed_sectors,
+      analysed_portfolio_absolute_value_change = .data$analysed_sectors_value_change,
+      total_portfolio_exposure = .data$portfolio_aum,
+      total_portfolio_percent_value_change = .data$portfolio_value_change_perc,
+      total_portfolio_absolute_value_change = .data$portfolio_value_change
     )
 
   portfolio_value_changes <- results_list$company_value_changes %>%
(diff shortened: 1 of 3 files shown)


MirijaHa:
FYI @jakubcervenka , please have a close look too as this affects documentation/output for users

R/wrangle_and_check.R
```diff
@@ -430,11 +478,42 @@ wrangle_results <- function(results_list, sensitivity_analysis_vars) {
...
+      .data$net_profits_baseline, .data$net_profits_ls,
+      .data$discounted_net_profit_baseline, .data$discounted_net_profit_ls,
+      !!!rlang::syms(sensitivity_analysis_vars)
+    ) %>%
+    dplyr::rename(
+      production_plan_company_technology = .data$plan_tech_prod,
+      # TODO: add once ADO3530 is merged
+      # direction_of_target = .data$direction,
+      production_baseline_scenario = .data$baseline,
+      production_target_scenario = .data$scen_to_follow_aligned,
```
MirijaHa, line 506: I wonder if itmakes sense to output also target scenario, since in practice we are only ever comparing baseline and shock scenario?

MirijaHa (commented):
1 minor questions, and you could style the package to get rid of a tiny nit, otherwise splendid

vignettes/articles/05-read-the-outputs.Rmd
```diff
@@ -194,14 +197,33 @@ loans (or alternatively of the credit limit) to a company.
...
+calculated as: `(pd_baseline + pd_change_shock) * lgd * exposure_at_default`.
 
 ### loans_company_pd_changes_annual_\<suffix\>
-To be continued.
+
+This file contains an overview on the company level of how annual probabilities
+of default change relative to the baseline scenario after the shock year.
+
+* `scenario_name`: Name that highlights the use of the Carbon budget.
+compensation mechanism and the Year of the introduction of a transition policy.
```
MirijaHa, line 212: typo nit year

vignettes/articles/05-read-the-outputs.Rmd
```diff
@@ -241,10 +279,66 @@ used.
...
-To be continued.
+
+This file contains an in detail overview of cash flow projections on the 
+company-technology level for the baseline and the shock scenarios.
+
+* `investor_name`: Name of the investor or bank for which the analysis was run.
+* `portfolio_name`: Name of the portfolio or loan book managed by the investor
+or bank for which the analysis was run.
+* `scenario_name`: Name that highlights the use of the Carbon budget.
+* `id`: internal id related to the company in the loan book or portfolio
```
MirijaHa, line 296: as id means something else for every asset type (and is NA for loans) I would encourage removing the columns

vignettes/articles/05-read-the-outputs.Rmd
```diff
@@ -241,10 +279,66 @@ used.
...
+* `volatility`: Volatility of the asset price of the company.
+* `production_unit`: Unit of the real production associated with the given
+combination of company and technology.
+* `price_unit_iea`: Unit of the market price for the given technology for IEA
+scenarios.
+* `price_unit_etr`: Unit of the market price for the given technology for ETR
+scenarios.
+* `price_baseline_scenario`: Market price for the given year and technology
+under the baseline scenario.
+* `baseline_source`: Source of the baseline scenario.
```
MirijaHa, line 332: this variable value will be "custom", which is maybe not a helpful value. Maybe it makes sense to reconsider once more if all returned columns here are helpful

MirijaHa (commented):
Did the first review before coffee and forgot to look at half of the code 🙈

vignettes/articles/05-read-the-outputs.Rmd
jacobvjk: will do for now although it could be helpful to assess coverage. But that is something we can think about at another point

vignettes/articles/05-read-the-outputs.Rmd
jacobvjk: we should think about that soon, I think we need to return a value somewhere that clarifies the inout scenarios. It's already in a follow up task on ADO, but yes, this is not helpful

R/wrangle_and_check.R
jacobvjk: I think it can be helpful to understand why the compensation behaves the way it does

R/wrangle_and_check.R
```diff
@@ -486,17 +486,16 @@ wrangle_results <- function(results_list, sensitivity_analysis_vars) {
   # company trajectories ----------------------------------------------------
   company_trajectories <- results_list$company_trajectories %>%
     dplyr::select(
-      .data$investor_name, .data$portfolio_name, .data$scenario_name, .data$id,
+      .data$investor_name, .data$portfolio_name, .data$scenario_name,
```
MirijaHa, line 489: I think id is now still referred to afterwards, since I get: ```
Adding missing grouping variables: `id` ```

jacobvjk: the issue was actually that the trajectory data is never ungrouped after the dcf calculation. I added that. Now it passes

MirijaHa (approved):
🕺

PR #334, opened 2022-02-16
P4B clarified in new prerequisites vignette

jakubcervenka:


README.md
```diff
@@ -22,6 +22,22 @@ data. For more information about the methodology and inquiries on
 running a pilot of the stress test in cooperation with 2DII, please
 contact <jacob@2degrees-investing.org>.
 
+## Prerequisites
```
MirijaHa, line 25: I would largely encourage removing this from the readme. The readme is developer facing, thus this information is self explanatory

MirijaHa (commented):
Please merge with master, as this branch is one commit behind.
Technically everything looks fine, on the content I would prefer that @jacobvjk reviews.
One remark though on the prerequesites copy pasted from P4B. At least the information that we only support  >= 3.5  is not correct, we support 3.4, on the remaining requirements I cannot say from the top of my head if they are correct.

README.md
MirijaHa: or rather remove points 1 und 2, point 3 makes sense also in a dev facing document.

jacobvjk: maybe we can just keep in point 3 and then extend the sentence below.. something like:
"We assume the user has successfully run at least the matching part of PACTA for Banks to produce the relevant project-specific raw input files. For a complete overview of the necessary prerequisites (including technical setup) please see [LINK]"

jacobvjk (commented):
@jakubcervenka do I understand correctly that the PACTA section of the vignette 00 is copied from PACTA material? In that case I am fine with the content. I do think the suggested change for the README would make it a bit nicer, but both feel free to comment

README.md
MirijaHa: sounds good to me

MirijaHa:
> @jakubcervenka do I understand correctly that the PACTA section of the vignette 00 is copied from PACTA material? In that case I am fine with the content. I do think the suggested change for the README would make it a bit nicer, but both feel free to comment

fine by me, just if keeping it please correct to 3.4

jacobvjk (approved):
Before merging the PR, please merge the current version of master again
otherwise lgtm

PR #335, opened 2022-02-17
remove hardcoded scenraio name

MirijaHa:
first in a number of micro PRs on the way to new price data. I copy pasted the comment from below, since I do not understand what is happening yet, though please suggest a better comment if you know on :).


jacobvjk (approved):
lgtm

PR #336, opened 2022-02-17
Unneeded cols

MirijaHa:
This PR reduces data set price data (wide) to the columns that are actually used to simplfiy switch to long data.


jacobvjk (approved):
lgtm

PR #341, opened 2022-02-21
Rm explicit exclusion from user workflow

jacobvjk:
closes ADO 1663: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1663

**This PR:**
- removes the explicit exclusion of companies from the transition risk stress test
- this never concerned the web tool, so no changes there
- for the user workflow, some specific functions for that functionality were removed
- exclusion-related arguments in other functions were also removed
- tests were updated
- test results change as expected (no overall changes except for one previously false exclusion, some minor trajectory changes for the previously excluded companies)
- The vignettes were updated

**NOTE:**
- there are some minor artefacts, e.g. an old configuration file I was unsure about. Happy to discuss if it can be removed
- also sorry this developed into a bit of a bigger PR.. there were more functions and files related to this than I initially thought


MirijaHa (changes requested):
1 issue, otherwise LGTM

MirijaHa (approved):
superbe

PR #343, opened 2022-02-22
Add model parameters for PA2022CH

cjyetman:
Adding the model parameters file for PA2022CH. To be honest, I don't know what any of these values mean/do, so I'd appreciate if @jacobvjk or someone that does would review them. I simply copied the PA2021NO file. Also, please let me know if there's anything else needed to do in this repo to prep for the PA2022CH launch. I'd like to have this stuff on online by the ned of Thursday at the latest. 🙏🏻 

One interesting thing I noticed.... I know we've had struggles before with the change in scenario names, especially I remember troubles with "NPS" being called "NPSRTS" in the 2002Q4 data. Here we also mention NPS explicitly... I wonder if that has any negative affect?


cjyetman:
closed in favor of #345

PR #348, opened 2022-02-23
Activate long pricedata

MirijaHa:
Using 2021Q1 data in long format. Impact on results is neglectable. FYI @jakubcervenka


jacobvjk (approved):
lgtm but requires follow up on file naming for long price data and scenarios expected in that file

PR #351, opened 2022-02-25
removing apostrophes in v04

jakubcervenka:
Very tiny fix: I have found some lost apostrophes in vignettes no04.


jacobvjk (approved):
lgtm

PR #357, opened 2022-03-16
filter out companies form pacta without valid exposures

jacobvjk:
removes rows from pacta results that do not have plan_carsten (company technology exposure) in the final year of the production forecast. These will not be usable when aggregating the results to the portfolio level later on

### DIFF
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -69,7 +69,9 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
 }
 
 #' Remove rows from PACTA results that belong to company-technology combinations
-#' for which there is no information on the exposure in the portfolio
+#' for which there is no information on the exposure in the portfolio. We join
+#' company results on the portfolio exposure based on the last available year of
+#' the production forecast. Hence we filter for missings in that year.
 #'
 #' @inheritParams calculate_annual_profits
 #' @inheritParams report_company_drops
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -85,9 +85,12 @@ remove_companies_with_missing_exposures <- function(data,
                                                     log_path) {
   n_companies_pre <- length(unique(data$company_name))
 
+  # we merge the exposure of the last year of the forecast on the company
+  # results for the aggregation, to get the closest picture to the shock year.
+  # Hence start_year + time_horizon
   companies_missing_exposure_value <- data %>%
     dplyr::filter(.data$year == .env$start_year + .env$time_horizon) %>%
-    dplyr::filter(is.na(.data$plan_carsten)) #%>%
+    dplyr::filter(is.na(.data$plan_carsten))
 
   data_filtered <- data %>%
     dplyr::anti_join(
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -1,6 +1,7 @@
 #' Process data of type indicated by function name
 #'
 #' @inheritParams run_stress_test
+#' @inheritParams report_company_drops
 #' @param data A tibble of data of type indicated by function name.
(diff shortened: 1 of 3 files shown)


R/process_data.R
```diff
@@ -62,6 +63,28 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
...
+  data <- data %>%
+    dplyr::group_by(
+      .data$investor_name, .data$portfolio_name, .data$equity_market,
+      .data$ald_sector, .data$technology, .data$scenario, .data$allocation,
+      .data$scenario_geography, .data$company_name
+    ) %>%
+    dplyr::arrange(.data$year) %>%
+    dplyr::mutate(
+      plan_carsten_missing = dplyr::if_else(
+        is.na(dplyr::last(.data$plan_carsten)),
```
MirijaHa, line 76: 1. I do not understand why checking for nas on dplyr::last is the relevant criterion here. Can u please add a comment
2. I would suggest adding a message/warning/logline, you name it,  on which companies are removed

jacobvjk: 1. re-wrote the function to hopefully clarify the logic
2. logging added

R/process_data.R
```diff
@@ -62,6 +68,59 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
...
+#' @return A tibble of data without rows with no exposure info
+#' @noRd
+remove_companies_with_missing_exposures <- function(data,
+                                                    start_year,
+                                                    time_horizon,
+                                                    log_path) {
+  n_companies_pre <- length(unique(data$company_name))
+
+  companies_missing_exposure_value <- data %>%
+    dplyr::filter(.data$year == .env$start_year + .env$time_horizon) %>%
```
MirijaHa, line 89: still i thinkt it would be a good idea to add a comment why you check for plan carsen on the last year

R/process_data.R
```diff
@@ -62,6 +68,59 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
...
+    dplyr::filter(is.na(.data$plan_carsten)) #%>%
+
+  data_filtered <- data %>%
+    dplyr::anti_join(
+      companies_missing_exposure_value,
+      by = c("company_name", "technology")
```
MirijaHa, line 95: there are 4 companies with missing exposure value. you observed the drops for 2 of the companies only, do you know what happened to the other 2

MirijaHa (commented):
2 small remarks

R/process_data.R
jacobvjk: yes, they have 0 production plans in the last year, while they have positive ones in the beginning, both for high carbon techs. They were previously removed as that is considered a technology phase out. So in both cases they are correctly not part of the aggregation

jacobvjk: the other 2 companies that previously slipped through, had positive production targets at the end of the forecast, hence they were not considered to phase out the technologies and in that case we must have an exposure.

R/process_data.R
jacobvjk: added

MirijaHa (commented):
last question, removing on company x tech, no company or compant x sector level is correct? Asking since we strive for complete sectors (i thought i commented that earlier but apparently did not..)

jacobvjk:
> last question, removing on company x tech, no company or compant x sector level is correct? Asking since we strive for complete sectors (i thought i commented that earlier but apparently did not..)

It's no problem for the calculation (and correctly so) if a company is only active in a subset of the technologies of a sector. However, it must be possible to get targets for all technologies of a sector at any point, because it is always possible that a company is active in all across all technologies.

MirijaHa (approved):
dazzling

PR #359, opened 2022-03-17
expose sector and tech to main function

MirijaHa:
mikro PR so that lower level function do not really on lookups for sector and tech anymore. Prep for Geo roleout.

### DIFF
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -186,9 +186,9 @@ process_capacity_factors_power <- function(data,
 process_price_data <- function(data, technologies, sectors, start_year, end_year,
                                scenarios_filter) {
   data_processed <- data %>%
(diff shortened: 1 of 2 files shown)


R/process_data.R
```diff
@@ -197,8 +197,8 @@ process_price_data <- function(data, technologies, sectors, start_year, end_year
       expected_levels_list =
         list(
           year = start_year:end_year,
-          ald_sector = sectors,
-          technology = technologies,
+          ald_sector = .env$sectors,
```
jacobvjk, line 200: I'm always a bit surprised by when to use `.env$` .. I usually get errors when using it outside a pipe.. but I'm guessing you did not get any messages

jacobvjk (approved):
lgtm

PR #361, opened 2022-03-18
Roleout geo

MirijaHa:
Allows using flexible scenario_geos. Allowed sectors are inferred from a mapper.


data-raw/scenario_geography_x_ald_sector.R
```diff
@@ -0,0 +1,53 @@
+# styler: off
+scenario_geography_x_ald_sector <- tibble::tribble(
```
jacobvjk, line 2: I would imagine we may have to extend this to add the scenario source at some point. Just as a pretext to what is likely to come in the future, obviously not needed now. But it is very unlikely that this mapping will e.g. also work for NGFS scenarios

jacobvjk (approved):
lgtm
works like a charm

PR #363, opened 2022-03-18
lbk prep aggregate plan_sec_carsten to sector not tech

jacobvjk:
sector level values in loan book preparation were previously erroneously aggregated to technology level
this led to wrong plan_sec_carsten values.
fixed the aggregation


MirijaHa (approved):
jsut to go sure, whereever else group_vars is used technology needs to be inlcuded?

jacobvjk:
> jsut to go sure, whereever else group_vars is used technology needs to be inlcuded?

yes, in the remaining places, it is being used correctly

PR #365, opened 2022-03-28
overview of all possible input parameters

jakubcervenka:
Both for
Scenario Geography
And all other input parameters

I thought this information is missing from the vignettes and can be included. It is more transparent and consistent.

### DIFF
--- a/vignettes/articles/03-run-stress-test.Rmd
+++ b/vignettes/articles/03-run-stress-test.Rmd
@@ -90,67 +90,39 @@ The meaning of each of the detail arguments is as follows:
 
 * `baseline_scenario` holds the name of the baseline scenario to be used in the
 stress test. We use a business-as-usual scenario for the baseline.
-  + Default set to: `NPS`
-  + Possible values: `NPS`
 * `shock_scenario` holds the name of the scenario that companies are forced to
 transition to via a late and sudden shock. We use an ambitious transition
 scenario as the shock scenario.
-  + Default set to: `SDS`
-  + Possible values: `SDS`
 * `lgd_senior_claims` is the loss given default used in the expected loss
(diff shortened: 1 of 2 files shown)


vignettes/articles/03-run-stress-test.Rmd
```diff
@@ -90,39 +90,67 @@ The meaning of each of the detail arguments is as follows:
```
MirijaHa, line 90: I would actually prefer to not explicitly put the defaults and allowed values in writing.. They change from time to time and thus we would always have to make sure to update both the single source of truth (stress_test_arguments) and the vignette. Also the information is already here https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/bf47f3895807654944f8334fc43bb38247e01fda/vignettes/articles/03-run-stress-test.Rmd#L85 
if you think it would be better to mention the defaults and possible values directly in the definitions we should programmatically get the value from stress_test_arguements, I can write a small helper function fot that.
On scenario_geographies the way it is presented here it is not apparent that the scenario_geographies are usually not fully supported but only for some sectors, so it might create unrealistic expectations. In that regard I would argue the explanation here is more adequate: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/bf47f3895807654944f8334fc43bb38247e01fda/vignettes/articles/02-prepare-loans-inputs.Rmd#L54

jakubcervenka: I take your advice. The list of geographies is now included in 02-prepare-loans-inputs.Rmd

I still think we should state the default values in the vignettes either in the list of input parameters or close and yes a programatic way of updating it would be great. I suggest we do this in a separate PR

MirijaHa (approved):
much clearer this way, thanks

PR #366, opened 2022-03-29
no env

MirijaHa:
Incorrectly used masking outside masking context. Apparently rlang < 1.0.0 did not throw an error in this case, so it was so far not detected.


maurolepore (approved):
LGTM

jakubcervenka:
Works now :))))))))))))

PR #368, opened 2022-03-30
avoid duplicates

MirijaHa:
To avoid that we do not run intro problems when this change https://github.com/2DegreesInvesting/r2dii.data/pull/270/files is releases on CRAN temporarily removing duplicates.


lindadelacombaz (approved):
Makes sense, LGTM

jdhoffa:
Thanks for being so on the ball @MirijaHa. 
I guess leave this open until `r2dii.data` is updated on CRAN then?

jdhoffa (approved):
Also makes sense to me! Thanks for being so quick @MirijaHa

Issue #307, opened 2022-01-12
Rename capacity_factors to capacity_factors_power


maurolepore:
Apparently capacity_factors (currently) only holds capacity_factors for the power sector ATM. This might be extended but I have no background knowledge on how likely and how soon that would happen, maybe @jacobvjk  knows more.
While the status quo is as it is I would suggest renaming read_capacity_factors to read_capacity_factors_power. If doing so please make sure it does not interfere with the webtool though/adjust webtool accordingly.

_Originally posted by @MirijaHa in https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/302#discussion_r782783448_

jacobvjk:
at this point, the power sector is the only sector that will require capacity factors. We get capacity on the asset/company level and need to translate that information to power generation by applying the capacity factor. This happens so that we can calculate the profits on units sold. For all other sectors that we currently cover, the asset level unit already holds the same type of information as is needed in the profit calculation.

PR #373, opened 2022-04-07
rename

MirijaHa:
this closes https://github.com/2DegreesInvesting/r2dii.climate.stress.test/issues/307
snapshot tests pass


tests/testthat/test-read_capacity_factors.R
```diff
@@ -1,15 +1,15 @@
...
           error", {
   testthat::expect_error(
-    read_capacity_factors(),
+    read_capacity_factors_power(),
     "Must provide 'path'"
   )
 })
 
 test_that("with all arguments set, old and new versions return data.frame", {
   test_path_new <- testthat::test_path("test_data", "capacity_factors_new.csv")
```
lindadelacombaz, line 10: since it is a csv file I am guessing that we do not have to rename it into capacity_factors_power_new.csv, but just wanted to pin this, just in case

MirijaHa: good observation, an you are right, the input files can have any name :)

lindadelacombaz (approved):

PR #377, opened 2022-04-08
replace plan carsten missings in start years with 0

jacobvjk:
ADO 4553

### DIFF
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -27,6 +27,10 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
       start_year = start_year,
       time_horizon = time_horizon
     ) %>%
+    set_initial_plan_carsten_missings_to_zero(
+      start_year = start_year,
+      time_horizon = time_horizon
+    ) %>%
     is_scenario_geography_in_pacta_results(scenario_geography_filter) %>%
     dplyr::filter(.data$investor_name == investor_name_placeholder) %>%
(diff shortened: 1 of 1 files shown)


R/process_data.R
```diff
@@ -133,6 +141,40 @@ remove_companies_with_missing_exposures <- function(data,
...
+set_initial_plan_carsten_missings_to_zero <- function(data,
+                                                      start_year,
+                                                      time_horizon) {
+  companies_missing_exposure_value <- data %>%
+    dplyr::filter(.data$year < .env$start_year + .env$time_horizon) %>%
+    dplyr::filter(is.na(.data$plan_carsten)) %>%
+    dplyr::mutate(plan_carsten = 0)
+
+  data_replaced <- data %>%
+    dplyr::filter(
```
MirijaHa, line 164: suggestion: how about instead of the filtering and rowbinding approach simply using mutate(if_else())

R/process_data.R
```diff
@@ -66,13 +69,18 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
       check_company_ticker_mapping()
   }
 
+  data_processed %>%
```
MirijaHa, line 72: turns out the fail when running snapshot test for loans is because column company_id is NA for loans. one could either create a placeholder company_id for loans or disregard that column in report_missings() for loans

MirijaHa (approved):
nice

PR #378, opened 2022-04-14
add linebreak

MirijaHa:
Purely aesthetic change to add a missing linebreak. You can see its effect when calculating the test equity results.


jacobvjk (approved):
yes!

PR #380, opened 2022-04-22
aggregate results accounting for sensitivity var

jacobvjk:
closes ADO4710: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/4710

This PR:
- adds the iteration variable on which a sensitivity analysis is run to the grouping vars in result aggregation
- handles the standard (non sensitivity) case

Expectation:
- Previously, ignoring the iteration variable in aggregating results lead to a situation, where only the lowest level results would differ between iteration runs, as these are non aggregated
- Any level of aggregation would lead to inflated sums or distorted means, because the aggregation to a higher level would e.g. not consider the need to keep the results of two different shock years apart from each other
- Test run shows this problem is now fixed

### DIFF
--- a/R/aggregate_results.R
+++ b/R/aggregate_results.R
@@ -14,9 +14,11 @@
 #' @param results_list A list of results.
 #' @param sensitivity_analysis_vars  String vector holding names of iteration
 #'   arguments.
+#' @param iter_var String vector holding the variable over which to iterate in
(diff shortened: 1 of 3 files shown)


MirijaHa (approved):
Snapshot tests also behaved as expected

PR #381, opened 2022-04-25
remove comp sector combination when no production given

jacobvjk:
closes ADO 4713 https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/4713

This PR:
- removes company-sector combinations in case they have 0 positive production in the relevant year of exposure (last year of the prod forecast)
- this catches the edge case that a company may produce HDVs but no LDVs and since both are part of the sector Automotive, that company will have a positive exposure but no production
- adds logging for which companies are removed due to this

Expectation:
- no such companies present in the trajectory results anymore
- no change to aggregate results since this case was previously caught there by chance

### DIFF
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -205,9 +205,6 @@ remove_sectors_with_missing_production <- function(data,
                                                    log_path) {
   n_companies_pre <- length(unique(data$company_name))
 
-  # we merge the exposure of the last year of the forecast on the company
-  # results for the aggregation, to get the closest picture to the shock year.
-  # Hence start_year + time_horizon
   companies_missing_sector_production <- data %>%
     dplyr::filter(.data$year == .env$start_year + .env$time_horizon) %>%
     dplyr::group_by(
--- a/R/process_data.R
+++ b/R/process_data.R
@@ -45,6 +45,11 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
       time_horizon = time_horizon,
       log_path = log_path
     ) %>%
+    remove_sectors_with_missing_production(
+      start_year = start_year,
+      time_horizon = time_horizon,
+      log_path = log_path
+    ) %>%
     stop_if_empty(data_name = "Pacta Results") %>%
     check_level_availability(
       data_name = "Pacta Results",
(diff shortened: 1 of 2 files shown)


jacobvjk:
From what I can tell, none of the trajectory results have NAs on the net profits anymore. The other expectations are also met

R/process_data.R
```diff
@@ -45,6 +45,11 @@ process_pacta_results <- function(data, start_year, end_year, time_horizon,
       time_horizon = time_horizon,
       log_path = log_path
     ) %>%
+    remove_sectors_with_missing_production(
```
MirijaHa, line 48: just to make sure, it makes more sense to put this removal after remove_companies_with_missing_exposures than before?

R/process_data.R
```diff
@@ -181,6 +186,73 @@ set_initial_plan_carsten_missings_to_zero <- function(data,
...
+#' covered in this analysis have any positive production. Such inconsistencies
+#' may arise e.g. because of unclear separation of the LDV and HDV sectors.
+#'
+#' @inheritParams calculate_annual_profits
+#' @inheritParams report_company_drops
+#' @param data tibble containing filtered PACTA results
+#'
+#' @return A tibble of data without rows with no exposure info
+#' @noRd
+remove_sectors_with_missing_production <- function(data,
```
MirijaHa, line 202: I would suggest to add unit tests, 1 where no companies are removed and 1 where companies are removed

R/process_data.R
```diff
@@ -181,6 +186,73 @@ set_initial_plan_carsten_missings_to_zero <- function(data,
...
+                                                   start_year,
+                                                   time_horizon,
+                                                   log_path) {
+  n_companies_pre <- length(unique(data$company_name))
+
+  # we merge the exposure of the last year of the forecast on the company
```
MirijaHa, line 208: 1. why we merge?
2. the closest picture to me sounds a bit like this is only approximately correct, just making sure that the formulation is intended to imply that the solution may not always work?

R/process_data.R
```diff
@@ -181,6 +186,73 @@ set_initial_plan_carsten_missings_to_zero <- function(data,
...
+    dplyr::summarise(
+      sector_prod = sum(.data$plan_tech_prod, na.rm = TRUE),
+      .groups = "drop"
+    ) %>%
+    dplyr::ungroup() %>%
+    dplyr::filter(.data$sector_prod <= 0)
+
+  data_filtered <- data %>%
+    dplyr::anti_join(
+      companies_missing_sector_production,
```
MirijaHa, line 226: robustness:
in l 214 you group by 6 varibales, in l 227 you merge by only 2 variables. This look incorrect from the pattern. E.g. you would remove rows for investor A, even if only investor B has the problem, or if only 1 scenario_geography is affected.

I would suggest to check if all the grouping is really neccessary and removing it if possible, if not please include the variables in merging, because otherwise this calls for future bugs. I would argue however, that if the problem exists in 1 scenario also the row for the other scenario should be removed.

MirijaHa (commented):
A few small comments.
A quick questions on the expectations and check, do I understand it correctly that you check ed that no output dataset expect for company trajectories changed?

R/process_data.R
jacobvjk: I would say so. Missing exposures is a more general case, this just cleans up one edge case that may slip through

R/process_data.R
jacobvjk: 1. the comment is not required here, I kept it from another function by accident. Will remove. The explanation is in the function description
2. obsolete because of (1), but nonetheless: we filter on the exposure in the last year of the forecast because that exposure is likely the most similar to what an exposure in the shock year will look like. We cannot say exactly what that exposure will look like though, so yes, this is an approximation. But it will always technically work

R/process_data.R
jacobvjk: okay, should be done now

MirijaHa: Just to go sure that it is really fine:

- did you verify that for all variables that you do not group by any longer there is some explicit filtering happning before?
- expectations: let us assume that companies_missing_sector_production holds a row for scenario A but not scenario B. is it correct behavior that in this case we would only remove the row for scenario A? I am nearly sure that it is not

MirijaHa (commented):
robustness again

jacobvjk:
1. explicit filtering for portfolio_name needs to be added in a follow up
2. I kept the scenario as a grouping and joining variable, since the production plans are the same across scenarios by definition so that they will only be removed for both or not at all

MirijaHa (approved):

PR #383, opened 2022-05-09
handle furhter edge cases to rm missings from comp profit trajectories

jacobvjk:
closes ADO 4120

This PR handles edge cases to remove missing values from the company annual profit calculation

Company-sector combinations, for which there is 0 production in the entire sector in the start year of the asset data are removed. This needs to be done because it is impossible to calculate SMSP based targets if there is no sector share to be kept constant.
new function `remove_sectors_with_missing_production_start_year()` in `process_pacta_results()`

Company-technology combinations, for which there is zero production in a high carbon technology across the full forecasting period of the asset data are also removed, since these technologies should be decreasing, but targets below zero are not possible.
new function `remove_high_carbon_tech_with_missing_production()` in `process_pacta_results()`

also renamed `remove_sectors_with_missing_production()` to `remove_sectors_with_missing_production_end_of_forecast()` for consistency

added tests for both new functions

This removed all remaining unexpected NAs

### DIFF
--- a/R/lookup.R
+++ b/R/lookup.R
@@ -93,6 +93,8 @@ p4b_scenarios_lookup <- c(
 # holds names of input arguments to run_stress_test that are not model parameters
 setup_vars_lookup <- c("input_path_project_agnostic", "input_path_project_specific", "output_path", "iter_var", "return_results")
 
+high_carbon_tech_lookup <- c("ICE", "Coal", "Oil", "Gas", "CoalCap", "GasCap", "OilCap")
+
 countries_for_regions_mapper_lookup <- tibble::tibble(
   region = c("brazil", "brazil", "india", "india", "japan", "japan", "russia", "russia", "south africa", "south africa", "united states", "united states"),
   isos = c("br", "br", "in", "in", "jp", "jp", "ru", "ru", "za", "za", "us", "us"),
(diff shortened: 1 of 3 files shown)


MirijaHa (commented):
I was so free to add 3 quick changes:

- moving a hardcoded vector to lookups
- merging on sector and technology to improve robustness for potential future added technologies
- fixed the not functioning logging for the high carbon case (since not all but only some rows per company may be removed using the diff on companies pre and post removal was not correct).


I also did some change checks:

for BONDS: remove_high_carbon_tech_with_missing_production removed some rows. Except for company_trajectories all results stayed the same.


for LOANS: Both new functions removed rows. All dataset had changed (different rownumbers and slight numerical changes in the means of the numeric variables).


If this is in line with your expectations I will approve the PR.

jacobvjk:
All changes you made, make sense to me and the checks behave as expected. The fact that there are slight differences in results for the case when a company gets removed that has 0 production in the start year is a bit of an unfortunate side-effect. This happens because the company may still have high carbon production built out after the start year which can be shocked normally. But since the SMSP cannot compensate the production in other technologies, I would say it is consistent with our target setting mechanism to remove the company. But I will add this as a methodological detail to revisit in the future

MirijaHa (approved):

PR #385, opened 2022-05-12
3112 terminal value

jacobvjk:
closes ADO 3112

**This PR:**
- adds an updated version of the terminal value calculation, as defined in 2DII's "limited visibility" report
- adds a function `calculate_terminal_value()` to the `calculate_annual_profits()` function
- adds a user facing parameter `growth_rate`, which determines the yearly growth of the last forecasted profit considered in the terminal value
- sets a default value of 0.03 for the growth rate, roughly in line with long term economic growth
- as the terminal value is added as an extra row in `end_year + 1` to the actual annual profits data, it is disregarded in the missingness checks, to avoid problems with the expected missingness pattern

**Expectations:**
- the company trajectory output will change: all rows pertaining to years <= end_year remain the same (excluding the added growth_rate_arg column), but an additional row for every combination of company, technology, scenario is added to hold the terminal value
- since the addition of the terminal value changes the net present values of the companies, we can expect that all other output files will change

Expectations are met for all assets


R/calculate.R
```diff
@@ -17,11 +17,14 @@
 #' @param start_year Numeric, holding start year of analysis.
 #' @param end_year Numeric, holding end year of analysis.
 #' @param time_horizon Considered timeframe for PACTA analysis.
+#' @param growth_rate Numeric, that holds the terminal growth rate of profits
+#'   beyond the final year in the DCF.
```
MirijaHa, line 21: please instead of final year mention `end_year` so it is obvious how args are related

R/calculate.R
```diff
@@ -152,3 +162,39 @@ calculate_exposure_by_technology_and_company <- function(asset_type,
 
   return(exposure_by_technology_and_company_filtered)
 }
+
+calculate_terminal_value <- function(data,
+                                     end_year,
+                                     growth_rate,
+                                     discount_rate) {
+  terminal_value <- data %>%
```
MirijaHa, line 170: how about a quick link to the paper that holds the formula

R/calculate.R
```diff
@@ -152,3 +162,39 @@ calculate_exposure_by_technology_and_company <- function(asset_type,
...
+    dplyr::filter(.data$year == .env$end_year) %>%
+    dplyr::group_by(
+      .data$investor_name, .data$portfolio_name, .data$id,
```
MirijaHa, line 173: please reconsider if really all grouping vars (e.g. phase_out) are needed

MirijaHa (changes requested):
Please also make sure to include validation of input argument growth rate (or I can do it to save you time). In particular plase note that the case where growth_rate == discount rate there would be a 0 divider in the formula.

R/calculate.R
jacobvjk: done

jacobvjk: done

R/calculate.R
MirijaHa: shouldnt that be resolvable by using mutate instead of summarise? 🤔

jacobvjk: yes, but then we would keep the production values from the end_year, which does not make sense from a model point of view. So I could do that, but then I would explicitly need to set some NA values here for the production fields

jacobvjk: done

jacobvjk: turns out summarizing leads to issues further down the line because the financial values were missing for the terminal value row. I ended up changing it to a mutate after all.

jacobvjk:
> Please also make sure to include validation of input argument growth rate (or I can do it to save you time). In particular plase note that the case where growth_rate == discount rate there would be a 0 divider in the formula.

Would you mind including the check @MirijaHa ? I might not have the time for it this week. In the first version I would say that discount_rate must be strictly larger than growth_rate and growth rate must be positive. Then we will cover most realistic cases and not have any mathematical issues

R/calculate.R
```diff
@@ -152,3 +162,40 @@ calculate_exposure_by_technology_and_company <- function(asset_type,
...
+      late_sudden = NA_real_,
+      scenario_change_aligned = NA_real_,
+
+    )
+
+  data <- data %>%
+    dplyr::bind_rows(terminal_value) %>%
+    dplyr::arrange(
+      .data$investor_name, .data$portfolio_name, .data$id,
+      .data$scenario_geography, .data$company_name, .data$ald_sector,
```
MirijaHa, line 196: As a last remark before I am approving it. Wenn comparing 2040 and 2041 it shows that in some cases info for 2041 is NA, and in some cases it is identical to 2040,  (e.g. proximity to target). It might be wise to revise per column if the way the columns are filled now is really correct and implement this TODO ```  # ADO 3112: ideally NPS and SDS (or more generally the columns with baseline
    # and target scenario names) should also get NA values. This makes no
    # difference for the rest of the analysis, but would be prettier.``` now, before creating confusion in the future

MirijaHa (commented):
last comment

R/calculate.R
jacobvjk: @MirijaHa sure, question is how best to get the scenario names without hard coding SDS and NPS again. But more than happy to include

MirijaHa: please hardcode them now so that the basis is correct and make it a problem for me to be resolved in https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_boards/board/t/Stress%20Testing%20Developers/Backlog%20items/?workitem=3129

jacobvjk: no need for a follow up, solved with sym()

MirijaHa (approved):
As the failing workflow is no related to the code changes feel free to merge

PR #389, opened 2022-05-19
Fix integration test

MirijaHa:
1) Snapshot tests were failing as argument for growth rate was not explicitly set, leading to an unallowed combination of discount and growth rate
2) fixing this showed a typo in the overview of allowed values for growth rate, fixed this.

FYI also @jakubcervenka


jacobvjk (approved):
noted, lgtm

PR #390, opened 2022-05-19
Fix masking

MirijaHa:
Embarrassing typo. Only merge after running snapshot tests against:  https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/389

The mistake was inconsequential in this case because data and .data in this case held the same data, however I thought I still bring it to your attention @maurolepore @lindadelacombaz and @julie-vienne just to raise awareness that such a mistake can happen.


jacobvjk (approved):
lgtm

PR #391, opened 2022-05-20
Vignette update: adding the terminal value growth rate to the list of parameters

jakubcervenka:


MirijaHa (commented):
For completeness, currently the <parameter>_arg columns are also documented in vignette 5, compare e.g. div_netprofit_prop_coef_arg . While I in general do not think that makes much sense I would suggest adding it for now for growth_rate_arg for consistency.

MirijaHa (approved):

PR #392, opened 2022-05-23
Update to the contact emails in the vignettes

jakubcervenka:
I have changed the contact email from jacob@.... to stresstest@....

I have added Mirja as the Author and Maintaner of the package to the DESCRIPTION
And the email to stresstest@...


jakubcervenka:
I am not sure if this is the correct way to modify DESCRIPTION, please let me know

MirijaHa:
something bad must have happened here, there is not diff: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/392/files 
I do not want to be included in the description but I like the idea of adding the stress test email

PR #394, opened 2022-05-24
Scenario flexibilisation

MirijaHa:
This closes 2b.
I will not merge this PR until the interface ist changed to accept source_x_scenario as inputs for baseline_scenario and shock_scenario, however wanted to single this functionality out to have a selfcontainted PR.
also fyi @jakubcervenka  since it affects user site docs


jakubcervenka (approved):
looks good

jacobvjk (approved):
I am still not 100% certain, if WEO2019 should be NPS instead of SPS, but given we name it this way in the data prep, I guess it is fine for now. May have to be adapted at some point, in case the scenario preparation gets refactored properly.

MirijaHa:
> I am still not 100% certain, if WEO2019 should be NPS instead of SPS, but given we name it this way in the data prep, I guess it is fine for now. May have to be adapted at some point, in case the scenario preparation gets refactored properly.

I am all for using the proper name NPS but do not want to mix more content into the PR.

Thanks for approval I will join it when the time is ripe

PR #397, opened 2022-05-25
remove portfolio level aggregation of results

jacobvjk:
closes ADO4858: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/4858

This PR:
- removes the aggregation of results to the company level (in `aggregate_results` and `wrangle_results`)
- this is required because each scenario only covers certain sectors
- if users need to run the stress test on different scenarios to cover all sectors they want to cover, they should not get conflicting statements on the impact on the total portfolio
- accordingly we only aggregate to the analysed sectors for each run.

Expectations:
- the only changes are that the columns `total_portfolio_exposure, total_portfolio_percent_value_change, total_portfolio_absolute_value_change` are removed from results `*_portfolio_value_changes_standard` and `*_company_value_changes_standard`. all else remains equal

Open:
- check expectations also hold for bonds and loans

### DIFF
--- a/R/aggregate_results.R
+++ b/R/aggregate_results.R
@@ -53,7 +53,7 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var)
     # aggregate value changes to sector level in given portfolio
     aggregate_value_change_to_sector_level(iter_var = iter_var) %>%
     # aggregate value changes to portfolio level and all covered sectors
-    aggregate_value_change_to_analysed_sectors_and_portfolio(iter_var = iter_var)
+    aggregate_value_change_to_analysed_sectors(iter_var = iter_var)
 
   company_value_changes <- company_value_changes %>%
     dplyr::select(-c(.data$plan_carsten, .data$plan_sec_carsten, .data$year))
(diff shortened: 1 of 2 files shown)


R/aggregate_results.R
```diff
@@ -207,14 +207,7 @@ aggregate_value_change_to_analysed_sectors_and_portfolio <- function(data,
       analysed_sectors_exposure = .data$asset_portfolio_value *
         sum(.data$plan_carsten, na.rm = TRUE),
       analysed_sectors_value_change = .data$analysed_sectors_exposure *
-        .data$VaR_analysed_sectors / 100,
-      portfolio_aum = .data$asset_portfolio_value,
```
MirijaHa, line 211: just to make sure before approving, are the columns that are no longer displayed to the users needed in earlier stages of the model? If not please make sure to delete their creation to declutter the code

MirijaHa (commented):
small question

R/aggregate_results.R
jacobvjk: I'm not sure I follow. The variables are not used before in the model, only for aggregation. I am removing them here. Where else in the code should I declutter @MirijaHa ?

MirijaHa: I was being imprecise. What I am referring to is for example ```.data$asset_portfolio_value```. If that variable is not needed for the model itself its creation could be deleted earlier on.

jacobvjk: `.data$asset_portfolio_value` is needed in the calculation of the exposure of the sum of all analysed sectors above. That section will remain in the analysis. The column itself also provides valuable information, so I would prefer to leave the remaining columns as they are.

MirijaHa (changes requested):
I think the vignette still needs to be updated to account for the changes

MirijaHa (approved):

jacobvjk:
> I think the vignette still needs to be updated to account for the changes

you are of course right :)

PR #398, opened 2022-05-31
No unneeded cap fac

MirijaHa:
This is another preparatory PR for https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/3129
It makes sure that capacity_factors are only imported, wrangled etc. if they are actually needed. To make this simpler I moved the conversion from cap to generation to data preprocessing, where it logically make also more sense IMO.
Snapshot tests passed (expectation: No changes)


jacobvjk (approved):
nice one

PR #401, opened 2022-06-08
rm uneeded function

MirijaHa:
super small PR to make diff for scenerio flex smaller. We do not need this function as this is already checked here: https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/3964bc57c67133e50197699d7412d03e4958dc9f/R/process_data.R#L519


jacobvjk (approved):
lgtm

PR #403, opened 2022-06-10
Enaable etp

MirijaHa:
This enables auto sector.
Note this currently only works, if the only sector is automotive. If it is possible that a scenario supports auto and another sector I have to rework the PR, otherwise I will do that in a follow up.


jacobvjk:
For now the assumption could work that a scenario will only cover automotive, but there will definitely be mixed scenarios in the future (GECO2020/1)

jacobvjk (approved):
works like a charm for now

MirijaHa:
> For now the assumption could work that a scenario will only cover automotive, but there will definitely be mixed scenarios in the future (GECO2020/1)

I reconsidered this ticket in the mean time and found a quick way to immediately activate also scenarios with both auto and not auto. I will close this PR an reopen another smaller one.

PR #406, opened 2022-07-15
Lrisk

jacobvjk:
creates a basic version of the LRISK model making use of the TRISK infrastructure

**summary:**
- new function `run_lrisk()` that follows the structure of `run_stress_test()`
- **where possible, functions from trisk are being reused**.
-  In some instances, functions from trisk are modified so they pass through information on the emission factors from the PACTA results
- In some instances the **behaviour under lrisk is so fundamentally different to trisk**, that **new functions** replace the ones from trisk (e.g. `set_litigation_trajectory()` instead of `set_late_sudden_trajectory()`)
- all **new parameters** were added as function args to `run_lrisk()` rather than extending the lookup. Some params that were already in the lookup from trisk are also being used here
- getting this workflow to run for loans required some heavy lifting in the `run_prep_calculation_loans()`, reading emission factors from the AR PAMS data set, as they are not a standard output in pacta for banks.
 
**note:**
- tests for run-stress-test currently fail, because they need to be adapted to using the new master data
- workflow runs through for all asset types
- workflow runs through with sensitivity analysis for all asset types

**open:**
- setting the trajectories needs another thorough step through to validate the behaviour
- the behaviour of the price trajectory should be discussed
- check why green technologies get a positive shock. This should not be the case imo, at least not due to the litigation cost and/or adjustments to the production pathway.  If this comes in through adjustments of the price or the emissions_factor, we should rethink if that behaviour is acceptable
- check why joining the AR PAMS data for emission factors in loan books leads to a loss of companies
- ...


R/calculate.R
```diff
@@ -39,6 +39,7 @@ calculate_annual_profits <- function(asset_type, input_data_list, scenario_to_fo
       start_analysis = start_year,
       end_analysis = end_year,
       time_frame = time_horizon,
+      baseline_scenario = baseline_scenario,
```
MirijaHa, line 42: dead argument

R/extend_scenario_trajectory.R
```diff
@@ -134,24 +134,36 @@ extend_scenario_trajectory_old <- function(data,
 #'   year of the analysis.
 #' @param time_frame Numeric. A vector of length 1 indicating the number of
 #'   years for which forward looking production data is considered.
+#' @param baseline_scenario Character. A vector of length 1 indicating baseline
```
MirijaHa, line 137: see above

MirijaHa (changes requested):
please investigate: Detected unexpected missings on variable production_change_target_scenario.

R/extend_scenario_trajectory.R
```diff
@@ -134,24 +134,36 @@ extend_scenario_trajectory_old <- function(data,
...
+#'   scenario
 #' @param target_scenario Character. A vector of length 1 indicating target
 #'   scenario
+#' @param emission_factors boolean. also summarise emmision factors? only needed
```
MirijaHa, line 141: for my understanding, are emmission factors only needed or only possible in the lrisk case? If it is the first I would suggest to also do it for trisk. It is not costly and saves some complication, a comment that it is only needed for lrisk would suffice

R/run_prep_calculation_loans.R
```diff
@@ -318,6 +328,9 @@ run_prep_calculation_loans <- function(input_path_project_specific,
 
   # ADO 1933 - for now, this only includes sectors with production pathways
   # in the future, sectors with emissions factors based pathways may follow
+  # tmsr <- scenario_data_market_share %>%
```
MirijaHa, line 331: no commented out code please (refers to all)

R/set_tech_trajectories.R
```diff
@@ -22,26 +22,35 @@
 #' @param scenario_to_follow_baseline Character. A string that indicates which
 #'   of the scenarios included in the analysis should be used to set the
 #'   baseline technology trajectories.
+#' @param emission_factors boolean. also summarise emmision factors? only needed
+#'   for lititgation risk at this point.
```
MirijaHa, line 26: same as above, if this can be around for trisk I would suggest not doing the distinguiation

R/set_tech_trajectories.R
```diff
@@ -430,3 +442,214 @@ filter_negative_late_and_sudden <- function(data_with_late_and_sudden, log_path)
...
+#'   end year of the analysis.
+#' @param analysis_time_frame Numeric. A vector of length 1 indicating the number
+#'   of years for which forward looking production data is considered.
+#'
+#' @return data frame
+set_litigation_trajectory <- function(data,
+                                      litigation_scenario,
+                                      shock_scenario,
+                                      litigation_scenario_aligned,
+                                      start_year = 2020,
```
MirijaHa, line 490: where do these hardcoded defaults come from and are they a good idea?

R/set_tech_trajectories.R
```diff
@@ -430,3 +442,214 @@ filter_negative_late_and_sudden <- function(data_with_late_and_sudden, log_path)
...
+                                      end_year = 2040,
+                                      analysis_time_frame = NULL,
+                                      log_path = NULL) {
+  analysis_time_frame %||% stop("Must provide input for 'time_frame'", call. = FALSE)
+
+  if (!"id" %in% names(data)) {
```
MirijaHa, line 496: ?

R/set_tech_trajectories.R
```diff
@@ -430,3 +442,214 @@ filter_negative_late_and_sudden <- function(data_with_late_and_sudden, log_path)
...
+    data$id <- "PortfolioLevel"
+  }
+  if (!"company_name" %in% names(data)) {
```
MirijaHa, line 499: ?

MirijaHa (changes requested):
please investigate: Detected unexpected missings on variable production_change_target_scenario. 

Also some small remarks

jacobvjk:
> please investigate: Detected unexpected missings on variable production_change_target_scenario.

fixed by removing superfluous variable after its only required use in extending the late sudden trajectory

R/set_tech_trajectories.R
jacobvjk: no, bad idea. removed

R/set_tech_trajectories.R
jacobvjk: removed

R/set_tech_trajectories.R
jacobvjk: removed

R/run_prep_calculation_loans.R
jacobvjk: removed

R/calculate.R
jacobvjk: removed

R/set_tech_trajectories.R
jacobvjk: removed distinction

R/extend_scenario_trajectory.R
jacobvjk: removed distinction

MirijaHa (commented):
when calculating equity results for shock_years 2025 and 2035 I lose different numbers of rows due to incompatibility with Merton Model. is this expected?

MirijaHa (changes requested):
Please fix fails in devtools::check

R/calculate_annual_pd_changes.R
```diff
@@ -59,6 +59,8 @@ calculate_pd_change_annual <- function(data,
     dplyr::group_by(
       .data$scenario_name, .data$scenario_geography, .data$investor_name,
       .data$portfolio_name, .data$id, .data$company_name, .data$ald_sector,
+      # TODO: clarify the appropriate summation for annual PD calculations
```
MirijaHa, line 62: still relevant?

R/calculate_annual_pd_changes.R
```diff
@@ -59,6 +59,8 @@ calculate_pd_change_annual <- function(data,
...
+      # .data$debt_equity_ratio, .data$volatility
```
MirijaHa, line 63: no commenting out pls, also would this not have an impact on trisk results?

R/run_prep_calculation_loans.R
```diff
@@ -326,6 +331,92 @@ run_prep_calculation_loans <- function(input_path_project_specific,
...
+      production_forecast_data %>% dplyr::distinct(.data$company_id, .data$name_company),
+      by = c("company_name" = "name_company")
+    ) %>%
+    dplyr::select(-.data$id) %>%
+    dplyr::rename(id = .data$company_id)
+
+  # This aggregates AR PAMS to company_technology level to get emission factors
+  # Note that this is done including all assets, i.e. we look at global
+  # emission factors here
+  # TODO: run for each scenario geography
```
MirijaHa, line 346: what are the consequences of not having this todo implemented?

R/run_prep_calculation_loans.R
```diff
@@ -326,6 +331,92 @@ run_prep_calculation_loans <- function(input_path_project_specific,
...
+      plan_emission_factor = stats::weighted.mean(
+        x = .data$ald_emissions_factor, w = .data$ald_production, na.rm = TRUE
+      ),
+      plan_tech_prod = sum(.data$ald_production, na.rm = TRUE),
+      .groups = "drop"
+    ) %>%
+    dplyr::ungroup()
+
+  # Join EFs based on PAMS
+  # FIXME: we are losing rows here, which we should not
```
MirijaHa, line 411: any more insights on this?

R/run_prep_calculation_loans.R
```diff
@@ -326,6 +331,92 @@ run_prep_calculation_loans <- function(input_path_project_specific,
...
+      ald_production = dplyr::if_else(.data$ald_production <= 0, 0, .data$ald_production),
+      ald_production_unit = as.character(.data$ald_production_unit),
+      ald_emissions_factor = as.numeric(.data$emissions_factor),
+      ald_emissions_factor_unit = as.character(.data$emissions_factor_unit)
+    ) %>%
+    dplyr::group_by(
+      .data$id, .data$company_name, .data$ald_sector, .data$technology, .data$year
+    ) %>%
+    dplyr::summarise(
+      # TODO: might introduce NaNs... check what to do here
```
MirijaHa, line 401: status?

R/run_prep_calculation_loans.R
```diff
@@ -98,6 +98,10 @@ run_prep_calculation_loans <- function(input_path_project_specific,
       sheet = "Company Indicators - PACTA Comp"
```
MirijaHa, line 98: In general: what are the expected changes on the results of loanbook prep? Did you check that there is no unintended impact on trisk (I would assume there is if we lose rows as indicated in one comment..)

MirijaHa (commented):
soem final comments, mostly on open TODOs

MirijaHa:
some further fishy things:

R/calculate_annual_pd_changes.R
jacobvjk: It would have an impact on trisk as well, yes

jacobvjk:
> some further fishy things:
> 
> * [2] "Component “duration_of_shock”: Mean relative difference: 0.09090909"

I adjusted the definition of the `duration_of_shock` in lrisk to the one used in trisk

R/calculate_annual_pd_changes.R
jacobvjk: not for this PR, but still relevant to clarify in general

R/calculate_annual_pd_changes.R
jacobvjk: Removed in favor of a more clarifying comment

R/run_prep_calculation_loans.R
jacobvjk: this was because of low carbon technologies that have zero buildout. They will not be in the asset based data, which is used to add EFs. These technologies need to remain in the loan book though, because the SMSP may require a buildout in such technologies, even if no such plans exist yet. I solved it with a left_join and subsequent imputation of zero value EFs for these cases.

R/run_prep_calculation_loans.R
jacobvjk: solved. Imputing technology type average values

R/run_prep_calculation_loans.R
jacobvjk: emissions factors can differ a bit on the regional level, as they are a weighted mean of the underlying technology types. If in one region different ways to extract oil are used compared to another region, the EFs may differ. This will not be large scale differences, but it would be more precise providing these on a regional level taking into account the local production mix

jacobvjk: Ideally this would be done when decoupling from pacta, as a similar aggregation will have to be done for the production data as well.

R/run_prep_calculation_loans.R
jacobvjk: After fixing the issue of why rows were dropped when adding emission factors, there are no further differences to results when running trisk on master versus this branch

MirijaHa: amazing

MirijaHa (approved):
🍰

PR #407, opened 2022-07-19
Crispy output 1

jacobvjk:
Create new output type that aggregates numbers to agreed upon CRISPY format
(see here: https://docs.google.com/spreadsheets/d/1vvvyBP4DLGv_XFylZ1Aln5Lcm3B78FmqDgvJNeIDSHo/edit#gid=0)

status:
- all the columns marked in green and yellow are currently provided
- the yellow one (ID) can be added once the decoupling from PACTA is prepared
- the orange one can only be added, once the stranded assets PR is properly integrated
- if lrisk results should get more output variables, these need to be added in a future PR
- full AR universe portfolio prepared as inputs here: https://github.com/2DegreesInvesting/r2dii.stress.test.data/pull/98

fyi @MirijaHa


jacobvjk:
(before and) after switching the loanbook prep to using _abcd in variable names instead of _ald, we still get odd warnings when running the loanbook preparation, along the lines of:

```
Warning message:
Expected 2 pieces. Additional pieces discarded in 20940 rows [13963, 13965, 13969, 13971, 13975, 13977, 13981, 13983, 13987, 13989, 13993, 13995, 13999, 14001, 14005, 14007, 14011, 14013, 14017, 14019, ...]. 
```

It seems like this might originate from a tidyr function, which I belive could be in the r2dii.analysis package. but it needs further investigation. In any case, results are produced.

jacobvjk:
the power sector seems to be large to handle in one iteration of the prep_calculation_loans function. Since loan book weights have no effect in CRISPY, it should be possibly to split the matched loan book into actually separate chunks of around 2000 loans each, which have shown to be handled well in other sectors

jacobvjk:
regarding the warning message mentioned above: This warning comes from r2dii.analysis::separate_metric_from_name(), which is a somewhat low level function. It would take some time to dive into the details of why this warning is thrown and what that means. Hence we should discuss if that is a priority

jacobvjk:
we currently lose a lot of companies, because we cannot map financial data to these. To solve this, we would need to update the prewrangled financial data to use 2021Q4 EO data. I will ignore that issue in this PR, but put it on the list of things to either do before I leave or describe the to do in detail

MirijaHa (changes requested):
please fix failing devtools::check()

R/calculate.R
```diff
@@ -44,6 +44,7 @@ calculate_annual_profits <- function(asset_type, input_data_list, scenario_to_fo
     set_baseline_trajectory(
       scenario_to_follow_baseline = scenario_to_follow_baseline
     ) %>%
+    # some companies dropped, document
```
MirijaHa, line 47: please do before merging

R/company_technology_asset_value_at_risk.R
```diff
@@ -8,10 +8,13 @@
...
 #' @param flat_multiplier Numeric. A ratio that determines for the asset type
 #'   if how strongly the DCF should propagate to value changes.
+#' @param crispy Boolean. Indicates if the output should be used for the CRISPY
+#'   database or for standard portfolio calculation (default).
 company_technology_asset_value_at_risk <- function(data,
                                                    shock_scenario = NULL,
                                                    div_netprofit_prop_coef = NULL,
-                                                   flat_multiplier = NULL) {
+                                                   flat_multiplier = NULL,
+                                                   crispy = FALSE) {
```
MirijaHa, line 17: same as with lrisk, can we simply always calculate this and save us the if statement?

R/format_loanbook_st.R
```diff
@@ -82,7 +82,7 @@ format_loanbook_st <- function(data,
       scenario = .data$metric,
       scenario_geography = .data$region,
       ald_sector = .data$sector,
-      company_name = .data$name_ald
+      company_name = .data$name_abcd
```
MirijaHa, line 85: what is the consequence on backward compatibility of this change?

R/wrangle_and_check.R
```diff
@@ -603,6 +604,20 @@ check_results <- function(wrangled_results_list, sensitivity_analysis_vars) {
       name_data = "Company Trajectories"
     )
 
+  # crispy results ----------------------------------------------------
+  wrangled_results_list$crispy_output %>%
+    report_missings(
+      name_data = "CRISPY Results"
+    ) %>%
+    report_all_duplicate_kinds(
+      composite_unique_cols = c(
```
MirijaHa, line 613: please move to lookup file

R/company_technology_asset_value_at_risk.R
MirijaHa: or actually let us leave it as is, this way it will be easier to switch of the other calculations once they are no longer needed.

MirijaHa (commented):
just some tiny comments

R/format_loanbook_st.R
jacobvjk: none, if we say that 2021Q4 data is a prerequisite for loan books. Which we do implicitly by the selection of files that are required for the loan book preparation. Otherwise users will have to rename said columns manually in the input data

R/wrangle_and_check.R
jacobvjk: just for my understanding, we should move it for crispy, but not for the other outputs, because crispy has a fixed set of cuc variables and the others do not?

MirijaHa:
> the power sector seems to be large to handle in one iteration of the prep_calculation_loans function. Since loan book weights have no effect in CRISPY, it should be possibly to split the matched loan book into actually separate chunks of around 2000 loans each, which have shown to be handled well in other sectors

Is this already reflected in the PR?

MirijaHa:
> (before and) after switching the loanbook prep to using _abcd in variable names instead of _ald, we still get odd warnings when running the loanbook preparation, along the lines of:
> 
> ```
> Warning message:
> Expected 2 pieces. Additional pieces discarded in 20940 rows [13963, 13965, 13969, 13971, 13975, 13977, 13981, 13983, 13987, 13989, 13993, 13995, 13999, 14001, 14005, 14007, 14011, 14013, 14017, 14019, ...]. 
> ```
> 
> It seems like this might originate from a tidyr function, which I belive could be in the r2dii.analysis package. but it needs further investigation. In any case, results are produced.

Can we close this?

MirijaHa:
> regarding the warning message mentioned above: This warning comes from r2dii.analysis::separate_metric_from_name(), which is a somewhat low level function. It would take some time to dive into the details of why this warning is thrown and what that means. Hence we should discuss if that is a priority

R/wrangle_and_check.R
MirijaHa: oh, i thought we had nicely put all of them into lookup, will close comment then

jacobvjk:
> > the power sector seems to be large to handle in one iteration of the prep_calculation_loans function. Since loan book weights have no effect in CRISPY, it should be possibly to split the matched loan book into actually separate chunks of around 2000 loans each, which have shown to be handled well in other sectors
> 
> Is this already reflected in the PR?

https://github.com/2DegreesInvesting/r2dii.stress.test.data/pull/98/files#diff-03a6383f51412a7b85ba22d15fb5a7379dc6a933465f9863088b634273e43086

jacobvjk:
> > (before and) after switching the loanbook prep to using _abcd in variable names instead of _ald, we still get odd warnings when running the loanbook preparation, along the lines of:
> > ```
> > Warning message:
> > Expected 2 pieces. Additional pieces discarded in 20940 rows [13963, 13965, 13969, 13971, 13975, 13977, 13981, 13983, 13987, 13989, 13993, 13995, 13999, 14001, 14005, 14007, 14011, 14013, 14017, 14019, ...]. 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems like this might originate from a tidyr function, which I belive could be in the r2dii.analysis package. but it needs further investigation. In any case, results are produced.
> 
> Can we close this?

yes, this is caught in process_scenarios

R/calculate.R
jacobvjk: standard company drop as before, from negative targets.

R/aggregate_results.R
```diff
@@ -76,12 +76,64 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var)
...
+        "shock_scenario_arg", "lgd_arg", "risk_free_rate_arg", "discount_rate_arg",
+        "growth_rate_arg", "div_netprofit_prop_coef_arg", "shock_year_arg",
+        "fallback_term_arg", "use_company_terms_arg"
+      )
+    )
+
+  crispy_output <- crispy_output %>%
+    dplyr::mutate(
+      roll_up_type = dplyr::if_else(
+        .data$asset_type_arg == "bonds",
```
MirijaHa, line 96: I would argue that a longer comment here would be very helpful. 
Also this would created duplicates in the db if we ran caluclations for equity and loans, since there is no more way left to discriminate between them ATM I am afraid

R/aggregate_results.R
```diff
@@ -76,12 +76,64 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var)
...
+      net_present_value_difference = .data$net_present_value_shock - .data$net_present_value_baseline,
+      pd_difference = .data$pd_shock - .data$pd_baseline
+    ) %>%
+    dplyr::select(
+      .data$company_name, .data$sector, .data$business_unit,
+      .data$roll_up_type, .data$scenario_geography, .data$baseline_scenario,
+      .data$shock_scenario, .data$lgd, .data$risk_free_rate, .data$discount_rate,
+      .data$dividend_rate, .data$growth_rate, .data$shock_year,
+      .data$net_present_value_baseline, .data$net_present_value_shock,
+      .data$net_present_value_difference, .data$term, .data$pd_baseline,
```
MirijaHa, line 126: nit: maybe but term before NPV like in the datapoitns doc

MirijaHa (changes requested):
1 bigger and 1 small aspect. Please also remove this check from snapshot tests, it will fail for crispy:  https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/d1fb16e519a5411f16e15292983c943daff54756/tests/testthat/test-run_stress_test.R#L24

R/aggregate_results.R
jacobvjk: I would argue we should turn it around in the doc, as the term refers to credit risk only and the NPV will not be affected by it

R/aggregate_results.R
jacobvjk: we can keep the asset type additionally for now and remove it, once the input is generated without the pacta step?

R/aggregate_results.R
MirijaHa: ok

R/aggregate_results.R
MirijaHa: sounds goodl. Maybe let us call it calculation_type actually..

jacobvjk: ok

R/aggregate_results.R
jacobvjk: done

jacobvjk:
> 1 bigger and 1 small aspect. Please also remove this check from snapshot tests, it will fail for crispy:
> 
> https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/d1fb16e519a5411f16e15292983c943daff54756/tests/testthat/test-run_stress_test.R#L24

test is adapted

MirijaHa (approved):
A bliss 🥳

PR #409, opened 2022-07-28
rm flat multiplier expcetion

MirijaHa:
This PR deactivates the flat multiplier by setting it to 1.


R/run_stress_test.R
```diff
@@ -182,7 +182,6 @@ read_and_process_and_calc <- function(args_list) {
 
   cat("-- Configuring analysis settings. \n")
 
-  flat_multiplier <- assign_flat_multiplier(asset_type = asset_type)
```
jacobvjk, line 185: after merging master, this will also have to be removed from `run_lrisk()`

R/run_stress_test.R
```diff
@@ -280,7 +279,7 @@ read_and_process_and_calc <- function(args_list) {
     company_technology_asset_value_at_risk(
       shock_scenario = transition_scenario,
       div_netprofit_prop_coef = div_netprofit_prop_coef,
-      flat_multiplier = flat_multiplier
```
jacobvjk, line 283: see above re `run_lrisk()`

jacobvjk (changes requested):
pls merge master and adapt `run_lrisk()` the same way as `run_stress_test()`
rest is great

R/run_stress_test.R
MirijaHa: done

jacobvjk (approved):
lgtm

PR #416, opened 2022-08-04
Decouple pacta

jacobvjk:
This PR decouples the stress test from pacta

**Status:**
The code runs through and generates the expected output files. I have clarified most points as well as was feasible in the remaining time. There are a few points open in the process_production_data function (commented out code that calls functions previously used on the pacta data). It should be checked on the production data, if these functions are needed in the same way, in an adapted form or not at all. I would recommend reading the function descriptions, running the stress test on the full aggregated production data and trying to identify if the cases laid out in the function description occur in the data

**Note:**
- for the snaps not to fail, a version of the aggregated production data would have to be read rather than trying to read pacta data. Changes are expected on all run_*risk() tests, as the inputs are completely different selections of companies
- running the stress test on the full aggregated production data is rather slow, so for development, i recommend using the `abcd_stress_test_input_test.csv` file instead, where appropriate (contains only one company).


R/convert_cap_to_generation.R
```diff
@@ -76,19 +76,19 @@ convert_cap_to_generation <- function(data,
 #'   translate company level power capacity to units sold. Contains information
 #'   on the technology (power sector only) and scenario_geography levels.
 #' @param baseline_scenario String holding name of baseline scenario.
+#' @param target_scenario String holding name of target scenario.
```
MirijaHa, line 79: why was target_scenario not needed here before?

R/extend_scenario_trajectory.R
```diff
@@ -176,12 +175,11 @@ extend_scenario_trajectory <- function(data,
   data <- data %>%
     dplyr::inner_join(
       scenario_data,
-      by = c("ald_sector", "technology", "scenario_geography", "scenario", "year")
```
MirijaHa, line 179: why can scenario be removed from merge vars?

R/read_company_terms.R
```diff
@@ -1,29 +0,0 @@
-#' Read in company terms
-#'
-#' Function reads in company terms and checks for existence of required columns.
-#'
-#' @inheritParams run_trisk
-#' @param path Path to dir holding company terms data.
-#'
-#' @return A tibble holding company terms data or NULL
-read_company_terms <- function(path, use_company_terms) {
```
MirijaHa, line 9: I think if we drop read_company_terms everything related to use_company_terms (https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/41f07ccf3254af18e7f0aa8fcb23ab98a2dbb874/R/run_trisk.R#L68) needs to disappear

R/read_production_data.R
```diff
@@ -0,0 +1,26 @@
...
+        scenario_geography = "c",
+        year = "d",
+        ald_sector = "c",
+        technology = "c",
+        plan_tech_prod = "d",
+        plan_emission_factor = "d",
+        plan_sec_prod = "d"
+      )
+    )
+
```
MirijaHa, line 24: please add validate_has_expected_cols for consistency

R/validate.R
```diff
@@ -8,10 +8,10 @@
 validate_input_values <- function(baseline_scenario, shock_scenario, scenario_geography,
                                   lgd, risk_free_rate, discount_rate, growth_rate,
                                   div_netprofit_prop_coef, shock_year,
-                                  fallback_term, use_company_terms, asset_type) {
```
MirijaHa, line 11: I would suggest to remove asset_type from stress_test_Arguments

data-raw/stress_test_arguments.R
```diff
@@ -12,7 +12,8 @@ stress_test_arguments <- tibble::tribble(
                "shock_year",    "double",   "2030",                     NA,     "2025",     "2035",
             "fallback_term",    "double",      "2",                     NA,        "1",        "5", # NOTE: before changing check cap_terms().
        "scenario_geography", "character", "Global", "AdvancedEconomies, Africa, AsiaPacific, Brazil, CentralandSouthAmerica, China, DevelopingEconomies, Emergingmarket&developingeconomies, EU, Eurasia, Europe, Global, India, Japan, LatinAmerica, MiddleEast, NonOECD, NorthAmerica, OECD, Russia, SouthAfrica, SoutheastAsia, US",         NA,         NA,
-        "use_company_terms",   "logical",  "FALSE",          "TRUE, FALSE",         NA,         NA
+        "use_company_terms",   "logical",  "FALSE",          "TRUE, FALSE",         NA,         NA,
+               "start_year",    "double",   "2021",                     NA,     "2021",     "2021"
```
MirijaHa, line 16: To my understanding there is only one possible value based on the data we have, or are there actually more options now we have gotten rid of pacta?

tests/testthat/test-run_trisk.R
```diff
@@ -53,20 +49,18 @@ test_that("with loans, without iteration, using minimum values of input argument
   expect_snapshot(lapply(results, as.data.frame))
 })
 
-test_that("with equity, without iteration, using maximum values of input arguments output is unchanged", {
+test_that("without iteration, using maximum values of input arguments output is unchanged", {
```
MirijaHa, line 52: I would suggest remoing this test and instead make the snapshot test for run_lrisk use the maxium values. Simply yo save time when running the tests.

R/process_data.R
```diff
@@ -658,23 +536,112 @@ st_process <- function(data, asset_type, fallback_term,
...
+    stop_if_empty(data_name = "Production Data") %>%
+    check_level_availability(
+      data_name = "Production Data",
+      expected_levels_list =
+        list(
+          year = start_year:(start_year + time_horizon),
+          scenario_geography = scenario_geography_filter
+        )
+    ) %>%
+    check_level_availability(
```
MirijaHa, line 619: why in 2 steps?

MirijaHa (commented):
Some things I noticed, mostly thing I do not understand

R/calc_scenario_prices.R
```diff
@@ -9,18 +9,18 @@
...
 #' @param baseline_scenario String holding name of the baseline scenario.
-#' @param shock_scenario String holding name of the shock scenario.
+#' @param target_scenario String holding name of the target scenario.
 #' @param transition_scenario Tibble with 1 row holding at least variables
 #'   `year_of_shock` and `duration_of_shock`.
 #' @param start_year Start_year of analysis
 #'
 #' @return A tibble holding late_and_sudden_prices
-calc_scenario_prices <- function(price_data, baseline_scenario, shock_scenario,
+calc_scenario_prices <- function(price_data, baseline_scenario, target_scenario,
```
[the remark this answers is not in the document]
jacobvjk: because we use confusing names for shock scenario and transition scenario later on. This way, it is more clearly disentangled.

MirijaHa (approved):

PR #420, opened 2022-08-15
included default and ranges for scc exp_share_damages_paid settlement…

frmfisc:
Included default and ranges for stress test arguments scc exp_share_damages settlement_factor 

Checks (as in validate.R) is currently problematic, as changes made affect not only lrisk but also trisk. 
To be precise, when running trisk, checking the lrisk stress test arguments should not occur, unsure how to implement that correctly (see line 23 and following)


R/validate.R
```diff
@@ -8,9 +8,15 @@
 validate_input_values <- function(baseline_scenario, shock_scenario, scenario_geography,
                                   lgd, risk_free_rate, discount_rate, growth_rate,
                                   div_netprofit_prop_coef, shock_year,
-                                  fallback_term, use_company_terms, asset_type) {
+                                  fallback_term, use_company_terms, asset_type,
+                                  settlement_factor = NULL, exp_share_damages_paid = NULL, scc = NULL) {
```
MirijaHa, line 12: As a reader of this function it can look a bit confusing, that for some arguments a default (NULL) is provided and for others it is not.
To make this very obvious how about:
1. adding a argument risk_type (currently with possible values "lrisk" and "trisk', make sure to add a check that the value is either of those)
2. in the documentation for settlement_factor etc. mention that values will be NULL for trisk
3. adapting the check in l 15 that the args settlement_factor etc are removed if risk_type is not 'lrisk'. it would also be a good idea to add the argument names explicitly in respective removal.

This solution may seem unnecessarily clumsy but it is saver, as it is currently implemented the function would e.g. not throw an error if we, incorrectly, do not provide values for settlement_factor etc. for lrisk.

R/validate.R
```diff
@@ -8,9 +8,15 @@
...
   input_args <- mget(names(formals()), sys.frame(sys.nframe()))
 
+  if(is.null(settlement_factor) & is.null(exp_share_damages_paid) & is.null(scc)) {
```
MirijaHa, line 15: nitpicking: there should be a lank between the "if" and the "(", please rerun the styler

MirijaHa (approved):
LGTM 🎆 
FYI @jakubcervenka this is ready for merge now :)

PR #421, opened 2022-08-30
Lrisk parameters to output

frmfisc:
Adjusted code such that
- lrisk parameters scc exp_share_damages_paid settlement_factor are in crispy output and 
- that more values can be imputed at the same time 
New variable needed to distinct lrisk and trisk

### DIFF
--- a/R/aggregate_results.R
+++ b/R/aggregate_results.R
@@ -34,7 +34,7 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var,
 
   if (risk_type == "lrisk") {
     lrisk_additional <- c("scc_arg", "settlement_factor_arg", "exp_share_damages_paid_arg")
-    merge_by_cols <- append(merge_by_cols, lrisk_additional)
+    merge_by_cols <- c(merge_by_cols, lrisk_additional)
   }
 
   crispy_output <- results_list$company_technology_npv %>%
@@ -109,7 +109,6 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var,
         .data$term, .data$pd_baseline, .data$pd_shock, .data$pd_difference
       )
   }
-
 
   return(list(
     company_trajectories = results_list$company_trajectories,
--- a/R/aggregate_results.R
+++ b/R/aggregate_results.R
@@ -30,7 +30,8 @@ aggregate_results <- function(results_list, sensitivity_analysis_vars, iter_var)
         "investor_name", "portfolio_name", "scenario_name", "scenario_geography",
         "company_name", "ald_sector", "asset_type_arg", "baseline_scenario_arg",
         "shock_scenario_arg", "lgd_arg", "risk_free_rate_arg", "discount_rate_arg",
-        "growth_rate_arg", "div_netprofit_prop_coef_arg", "shock_year_arg",
+        "growth_rate_arg", "scc_arg", "settlement_factor_arg", "exp_share_damages_paid_arg",
+        "div_netprofit_prop_coef_arg", "shock_year_arg",
         "fallback_term_arg", "use_company_terms_arg"
       )
     )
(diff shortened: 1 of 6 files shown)


MirijaHa:
Dear @frmfisc  Snapshot tests fail for LRISK but not for TRISK. This is in line with expecations, as your changes should only affect LRISK output. 
Can u please describe which expectation you have an changes in the lrisk output and if you verified that the expectations hold true?

R/aggregate_results.R
```diff
@@ -16,63 +16,121 @@
...
-  crispy_output <- results_list$company_technology_npv %>%
-    dplyr::inner_join(
-      results_list$company_pd_changes_overall,
-      by = c(
-        "investor_name", "portfolio_name", "scenario_name", "scenario_geography",
-        "company_name", "ald_sector", "asset_type_arg", "baseline_scenario_arg",
-        "shock_scenario_arg", "lgd_arg", "risk_free_rate_arg", "discount_rate_arg",
-        "growth_rate_arg", "div_netprofit_prop_coef_arg", "shock_year_arg",
-        "fallback_term_arg", "use_company_terms_arg"
+  if (risk_type == "trisk") {
```
MirijaHa, line 27: Your code is all correct but induces a lot of duplicates code, which makes it a bit harder to read/to maintain. 

To improve this I would suggest that for all behaviors that depend on risk type you only put the aspect that differs into an if-else logic, or even just an if logic.

E.g. here you could:
1. create a vector merge_by_cols <- ("investor_name", "portfolio_name",...) for trisk
2. append the new variables to that vector if risk_type is "lrisk"
3.  a pass this vector to by in the inner_join. 

Similar simplifications could be made also for the other if else logics 
(tipp: You will sometimes have to break the pipes to simplify the code)

MirijaHa (commented):
LGTM overall, before I approve please check out the comments :)

R/aggregate_results.R
frmfisc: hope this is addressed

frmfisc:
I have some questions on the snapshots, 
a) how do i see that they fail?
b) what exactly do they show? if it is the crispyoutput i expect lrisk to have 3 columns more

MirijaHa:
> I have some questions on the snapshots, a) how do i see that they fail? b) what exactly do they show? if it is the crispyoutput i expect lrisk to have 3 columns more

a) running devtools::test will show fails for respective tests
b) So that means: no changes for trisk, 3 new columns for lrisk but no changes otherwise?

R/wrangle_and_check.R
```diff
@@ -288,19 +289,30 @@ check_results <- function(wrangled_results_list, sensitivity_analysis_vars) {
...
-      composite_unique_cols = c(
-        "company_name", "sector", "business_unit", "roll_up_type",
-        "scenario_geography", "calculation_type", "baseline_scenario",
-        "shock_scenario", "lgd", "risk_free_rate", "discount_rate",
-        "dividend_rate", "growth_rate", "shock_year", "term"
-      )
+      composite_unique_cols = composite_unique_cols_crispy_results
     )
 
+
```
MirijaHa, line 313: nit: no unneeded whitespace pls

R/wrangle_and_check.R
```diff
@@ -288,19 +289,30 @@ check_results <- function(wrangled_results_list, sensitivity_analysis_vars) {
...
+  composite_unique_cols_crispy_results <- c(
+    "company_name", "sector", "business_unit", "roll_up_type",
+    "scenario_geography", "calculation_type", "baseline_scenario",
+    "shock_scenario", "lgd", "risk_free_rate", "discount_rate",
+    "dividend_rate", "growth_rate", "shock_year", "term"
+  )
+
+  if (risk_type == "lrisk") {
+    composite_unique_cols_lrisk <- c("scc", "settlement_factor", "exp_share_damages_paid")
```
MirijaHa, line 301: super clear and readible 👍

MirijaHa (commented):
LGTM, just a tiny nit before I approve.

R/aggregate_results.R
```diff
@@ -16,23 +16,31 @@
...
+    "investor_name", "portfolio_name", "scenario_name", "scenario_geography",
+    "company_name", "ald_sector", "asset_type_arg", "baseline_scenario_arg",
+    "shock_scenario_arg", "lgd_arg", "risk_free_rate_arg", "discount_rate_arg",
+    "growth_rate_arg", "div_netprofit_prop_coef_arg", "shock_year_arg",
+    "fallback_term_arg", "use_company_terms_arg"
+  )
+
+  if (risk_type == "lrisk") {
+    lrisk_additional <- c("scc_arg", "settlement_factor_arg", "exp_share_damages_paid_arg")
+    merge_by_cols <- append(merge_by_cols, lrisk_additional)
```
MirijaHa, line 37: one more tiny nit: instead of append you can simply use c to combine to vectors, append is more typically used for lists.

MirijaHa (commented):
added one more nit while at it 😄

MirijaHa (approved):
🥳

PR #432, opened 2022-10-19
Weo 2019

frmfisc:
i am snaptest confused - i suppose the path problem is related to that.


MirijaHa:
> i am snaptest confused - i suppose the path problem is related to that.

Can u please elaborate on your confusion?

MirijaHa (approved):

PR #437, opened 2022-11-29
Geco2021

frmfisc:
update geco2021


tests/testthat/_snaps/stress_test_arguments.md
```diff
@@ -1,23 +1,13 @@
...
-       7 growth_rate             double    0.03          <NA>              0.01  0.099
-       8 div_netprofit_prop_coef double    1             <NA>              0.8   1    
-       9 shock_year              double    2030          <NA>              2025  2035 
-      10 scenario_geography      character Global        Africa, AsiaPaci~ <NA>  <NA> 
-      11 start_year              double    2021          <NA>              2021  2021 
-      12 settlement_factor       double    1             <NA>              0     1    
-      13 exp_share_damages_paid  double    0.027         <NA>              0     1    
-      14 scc                     double    40            <NA>              0     10000
+      tibble [14 x 6] (S3: tbl_df/tbl/data.frame)
+       $ name   : chr [1:14] "asset_type" "baseline_scenario" "shock_scenario" "lgd" ...
```
MirijaHa, line 7: unfortunately this sort of made id worse.. you can see from the ... that it is still truncating, this time rows, and I would assume that is evene worse... How about removing this change for now (use git revert if you want to be fancy), merging the PR and looking at the other options that dear @maurolepore  suggested in a follow up?

PR #440, opened 2023-01-06
Add carbon price file

frmfisc:
This is the first step in integrating the carbon tax 

this pr only seeks to add the dataset into the stress test nothing else. 
snap tests remained unchanged as discussed and expected 

thanks for the review


MirijaHa (approved):
LGTM, just to go sure, you plan do introduce processing of that file in the next pr (compare all the process.. -functions) or is there no need for that?

PR #441, opened 2023-01-09
Integration carbon tax

frmfisc:
this pr includes the carbon tax within the net profit equation for the trisk model
in a next step carbon tax, in the figure of different models, will be added as input option for the user 

as only trisk has carbon tax, i used the old function, renamed to calculate_netprofits_lrisk, for lrisk. reason for not doing it in an ifelse is that i fear the net_profits_calcuation function will be very dense (there are some if statements planned) - not sure of course that this is correct

snap test change for trisk, which is expected 

there remains one error when running devtools::check() which for the life of me i cannot backtrace let alone fix 

i hope it is not too confusing - thanks for the review.


MirijaHa (changes requested):
I assume the problem you are referring to is the error in devtools::check() that comes from the unit tests.
If you run devtools::test() you will spot this error 
```
(test-run_trisk.R:38): without iteration, using maximum values of input arguments output is unchanged
<dplyr:::mutate_error/rlang_error/error/condition>
Error in `dplyr::mutate(., t_calc = seq(0, (dplyr::n() - 1)), discounted_net_profit_baseline = .data$net_profits_baseline/(1 + 
    .env$discount_rate)^.data$t_calc, discounted_net_profit_ls = .data$net_profits_ls/(1 + 
    .env$discount_rate)^.data$t_calc)`: Problem while computing `t_calc = seq(0, (dplyr::n() - 1))`.
x `t_calc` must be size 0 or 1, not 2.
```

In order to fix this:
- find the part in the sourcecode
- put a browser
- run the respective test in test-run_trisk
- enjoy

R/financial_functions.R
```diff
@@ -16,7 +16,37 @@
...
+        .data$late_sudden * .data$late_sudden_price * .data$net_profit_margin -
+          .data$production_compensation * .data$late_sudden_price * .data$net_profit_margin * (1 - .data$proximity_to_target)
+        # TODO: ADO4109 - should the market size penalty only be applied to laggards?
+
+      )
+    )
+
+}
+
+calculate_net_profits_lrisk <- function(data) {
```
MirijaHa, line 49: You might want to change the naming if you want to be very good. The difference per se between the functions is not that one is for lrisk, but that one calculates net profits with carbon tax and one without.

R/financial_functions.R
```diff
@@ -16,7 +16,37 @@
...
+#' @param shock_year A numeric vector of length one that indicates in which year
+#'   the policy shock strikes in a given scenario.
+#' @param end_year Numeric, holding end year of analysis.
+#' @param carbon_data NGFS carbon prices.
+calculate_net_profits <- function(data,
+                                  shock_year,
+                                  end_year,
+                                  carbon_data) {
+
+  carbon_data <- carbon_data %>% dplyr::filter(.data$model == "MESSAGEix-GLOBIOM 1.0") %>% dplyr::select(-c(scenario_geography))
```
MirijaHa, line 28: shouldnt this happen in process_data? 
Also why is geo in the data when it is deselected? 
Also please do not hard code in the code (e.g. MESSAGEix-GLOBIOM 1.0), if it is not evitable please add it as a lookup.R, compare e.g. end_year_lookup  :)

R/financial_functions.R
```diff
@@ -16,7 +16,37 @@
...
+    dplyr::mutate(
+      carbon_tax = ifelse (year > shock_year, .data$carbon_tax ,0),
+      production_compensation = .data$late_sudden - .data$baseline,
+      net_profits_baseline = .data$baseline * .data$Baseline_price * .data$net_profit_margin,
+      net_profits_ls = dplyr::if_else(
+        .data$direction == "declining",
+        .data$late_sudden * (.data$late_sudden_price - (1-market_passthrough) * .data$carbon_tax* .data$emission_factor) * .data$net_profit_margin,
+        .data$late_sudden * .data$late_sudden_price * .data$net_profit_margin -
+          .data$production_compensation * .data$late_sudden_price * .data$net_profit_margin * (1 - .data$proximity_to_target)
+        # TODO: ADO4109 - should the market size penalty only be applied to laggards?
```
MirijaHa, line 42: comment still relevant? If not please drop

R/financial_functions.R
```diff
@@ -16,7 +16,37 @@
...
+#' @param carbon_data NGFS carbon prices.
+calculate_net_profits <- function(data,
+                                  shock_year,
+                                  end_year,
+                                  carbon_data) {
+
+  carbon_data <- carbon_data %>% dplyr::filter(.data$model == "MESSAGEix-GLOBIOM 1.0") %>% dplyr::select(-c(scenario_geography))
+  market_passthrough <- 0
+
+  data <-data %>%
```
MirijaHa, line 31: i would suggest to run the styler before merge the pr to get rid of these tiny formatting issues

R/process_data.R
```diff
@@ -335,6 +335,21 @@ process_scenario_data <- function(data, start_year, end_year, sectors, technolog
   return(data_processed)
 }
 
+#' Process data of type indicated by function name
+#'
+#' @inheritParams process_carbon_data
```
MirijaHa, line 340: this is not entirely correct, if you are using inheritParams you need to inherit from a function where the arguments are described

R/process_data.R
```diff
@@ -335,6 +335,21 @@ process_scenario_data <- function(data, start_year, end_year, sectors, technolog
...
+#'
+#' @return A tibble of data as indicated by function name.
+#' @noRd
+process_carbon_data <- function(data, start_year, end_year) {
+  data_processed <- data  %>%
+    dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$end_year)) %>%
+    #dplyr::filter(.data$model %in% .env$model_filter)
```
MirijaHa, line 347: no commented out code pls

tests/testthat/test-financial_functions.R
```diff
@@ -1,12 +1,28 @@
 test_that("calculate_net_profits penalizes companies for late build out of low
           carbon technologies", {
   input_data <- tibble::tribble(
-    ~company_name, ~baseline, ~late_sudden, ~Baseline_price, ~late_sudden_price, ~net_profit_margin, ~direction, ~proximity_to_target,
-    "leader", 100, 150, 10, 10, 0.1, "increasing", 1,
-    "laggard", 100, 150, 10, 10, 0.1, "increasing", 0.5
+    ~company_name, ~baseline, ~late_sudden, ~Baseline_price, ~late_sudden_price, ~net_profit_margin, ~direction, ~proximity_to_target, ~scenario_geography, ~year, ~emission_factor,
```
MirijaHa, line 4: So far you only adapted the test so they still run but did not add a new one to test specifically the behavior of the carbon tax.
How about you add a unit test where you calculate net_profits on the same input data with or without carbon tax and make some assuptions on the differences (happy to explain in more detail)

MirijaHa (commented):
Nothing breaking, just some well meaning suggestions

R/financial_functions.R
frmfisc: ok makes sense ill adjust it

R/financial_functions.R
frmfisc: well the thing is i now wanna integrate carbon tax as a user defined input, i.e. they will be able to select the model they want to use so i will get rid of the hardcoded value now in the next step. is that ok and sensible? geography is in there bc i intend to have a better dataset at some point that contains different carbon prices for different regions

R/financial_functions.R
frmfisc: still relevant in fact but ill keep it in mind and try to solve it with the team

tests/testthat/test-financial_functions.R
frmfisc: that would be amazing!

R/financial_functions.R
frmfisc: done

R/process_data.R
frmfisc: understood

R/process_data.R
MirijaHa: you use inhertigParams to inherit the parameter definitions form another function that has the paramters defined (in concrete another fucntion that has data, start_year and end_year defined). You cannot inherit data here, since it is specific to process_carbon_data. The other 2 params you could e.g. inherit from : process_production_data

MirijaHa (approved):

PR #444, opened 2023-01-17
No carbon tax break net profits

frmfisc:


R/financial_functions.R
```diff
@@ -45,23 +45,75 @@ calculate_net_profits <- function(data,
     )
 }
 
+#' Calculates annual net profits on the company-technology level for the
+#' baseline and late and sudden scenarios without carbon tax Climate laggards which need to build
+#' out their production in increasing technologies to compensate for their
```
MirijaHa, line 50: 1. please use the roxygen sceletons to add documentation to make sure that they are formatted corretly and complete. Simply click on a function and click Insert Roxygen Sceleton under the Code pain
2. I would suggest to put the details not in calculate_net_profits_without_carbon_tax (here it is enought to say that it calculates the net profits for the following cases) and instead go into detail on the smaller functions.

R/financial_functions.R
```diff
@@ -45,23 +45,75 @@ calculate_net_profits <- function(data,
...
+#'
+#' @param data A data frame containing the production forecasts of companies
+#'   under baseline and late and sudden, market prices/costs, company net profit
+#'   margins, the proximity to target in the production forecast period and an
+#'   indication of the direction of the technology.
+#'   @param direction either declining or increasing
+
 calculate_net_profits_without_carbon_tax <- function(data) {
+  baseline <- calculate_net_profits_baseline(data)
+  shock_increasing_technologies <- calculate_net_profits_shock_increasing_technologies(data = data %>% dplyr::filter(direction == "increasing"))
```
MirijaHa, line 70: did you run styler already?

R/financial_functions.R
```diff
@@ -45,23 +45,75 @@ calculate_net_profits <- function(data,
...
+
+#' Calculates annual net profits on the company-technology level for the
+#' baseline
+#' @param data A data frame containing the production forecasts of companies
+#'   under baseline and late and sudden, market prices/costs, company net profit
+#'   margins, the proximity to target in the production forecast period and an
+#'   indication of the direction of the technology.
+
+calculate_net_profits_baseline <- function(data) {
+  data <- data %>%
```
MirijaHa, line 85: I would suggest to explicitly return from functions, i.e. put return(data) at the end. It is not always needed, but there is some rare cases where not returning explicitly will return the wrong thing..

R/financial_functions.R
```diff
@@ -45,23 +45,75 @@ calculate_net_profits <- function(data,
...
+#'   indication of the direction of the technology.
+#'   @param direction either declining or increasing
+
 calculate_net_profits_without_carbon_tax <- function(data) {
+  baseline <- calculate_net_profits_baseline(data)
+  shock_increasing_technologies <- calculate_net_profits_shock_increasing_technologies(data = data %>% dplyr::filter(direction == "increasing"))
+  shock_declining_technologies <- calculate_net_profits_shock_declining_technologies(data = data %>% dplyr::filter(direction == "declining"))
+
+  data_test <- dplyr::full_join(shock_increasing_technologies, shock_declining_technologies)
+  data_test <- dplyr::full_join(data_test, baseline)
```
MirijaHa, line 74: data_test is not the nicest name

MirijaHa (commented):
Mostly just formalities. I think that the code is already much easier to understand thanks to your refacotoring 🚀 

This would be a good point in time to add unit tests for the functions 🤓

MirijaHa (commented):
please also fix the new note (no global binding for direction)

R/financial_functions.R
frmfisc: done

R/financial_functions.R
frmfisc: changed it back to data

R/financial_functions.R
frmfisc: 2. makes sense, got it. 1. still needs to be addressed

R/financial_functions.R
frmfisc: actually yes

R/financial_functions.R
frmfisc: 1. done

PR #445, opened 2023-01-23
Break down net profits carbon tax

frmfisc:
this branch breaks down the net profit equation - with a carbon tax included 
- new unit test 
- rewriting equation 
- put the processing bit i.e. selection of a model and scenario in the right place


MirijaHa (changes requested):
Please fix the new notes, errors etc and then rerequest review :)

tests/testthat/test-financial_functions.R
```diff
@@ -96,6 +92,95 @@ test_that("calculate_net_profits does not apply penalty on lost profits for high
   )
 })
 
+test_that("calculate_net_profits does not apply carbon tax on high
```
MirijaHa, line 95: nice selection of test cases :)

R/process_data.R
```diff
@@ -344,6 +344,8 @@ process_scenario_data <- function(data, start_year, end_year, sectors, technolog
 process_carbon_data <- function(data, start_year, end_year) {
   data_processed <- data %>%
     dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$end_year)) %>%
+      dplyr::filter(.data$model == "MESSAGEix-GLOBIOM 1.0") %>%
+      dplyr::select(-c(scenario_geography))  %>%
```
MirijaHa, line 348: why are u deselecting scenario_geogrpahy? It seems there is not filtering by it, so why include it in the first place? 🤔

MirijaHa (approved):
just one small question for my understanding, otherwise lgtm :)

PR #447, opened 2023-02-03
Input args carbon tax

frmfisc:
this PR adds carbon tax model to the stress test args 
adjustements have to be made as lrisk does not have these new inputs 
carbon price not read in for lrisk 

Thanks!


MirijaHa (commented):
Can you please explain briefly, why the snapshot tests for trisk fail? :)

tests/testthat/test-validate.R
```diff
@@ -2,7 +2,7 @@
 test_that("Error is thrown if input values are of incorrect type", {
 
   # numeric
-  expect_error(validate_input_values(
+   expect_error(validate_input_values(
```
MirijaHa, line 5: nit: please tun automatic styler for every PR, this looks like an accidentally introduced whitespace

R/read.R
```diff
@@ -1,4 +1,4 @@
-st_read_agnostic <- function(dir, start_year, sectors) {
+st_read_agnostic <- function(dir, start_year, sectors, risktype) {
```
MirijaHa, line 1: nit: You call this variable risk_type when calling validate_input_values, I would suggest to stick to the name

R/read.R
```diff
@@ -7,14 +7,25 @@ st_read_agnostic <- function(dir, start_year, sectors) {
...
   }
 
+  if (risktype == "trisk") {
   out <- list(
     capacity_factors_power = capacity_factors_power,
     df_price = read_price_data(price_data_file(dir)),
     scenario_data = read_scenario_data(scenario_data_file(dir, start_year)),
     financial_data = read_financial_data(financial_data_file(dir)),
     production_data = read_production_data(production_data_file(dir)),
     carbon_data = read_carbon_data(carbon_price_data_file(dir))
```
MirijaHa, line 17: I would suggest to make the if else condition only around carbob_data and also add a comment why this is happening. This would make the difference much easier to spot and understand

R/process_data.R
```diff
@@ -341,13 +341,21 @@ process_scenario_data <- function(data, start_year, end_year, sectors, technolog
...
-    dplyr::filter(dplyr::between(.data$year, .env$start_year, .env$end_year)) %>%
-      dplyr::filter(.data$model == "MESSAGEix-GLOBIOM 1.0") %>%
-      dplyr::select(-c(scenario_geography))  %>%
-    # dplyr::filter(.data$model %in% .env$model_filter)
-    stop_if_empty(data_name = "Carbon Data")
+process_carbon_data <- function(data, start_year, end_year, carbon_price_model) {
+
+  data_processed <- data
+
+  if (is.null(data_processed)) {
```
MirijaHa, line 348: please add comment that explains when data_process is null

R/validate.R
```diff
@@ -3,12 +3,14 @@
 #' Checks that user inputs are within defined ranges.
 #'
 #' @inheritParams run_lrisk
+#' @param carbon_price_model Character vector, indicating which NGFS model is used in regards to
+#'   carbon prices. Default is no carbon tax.
 #' @param risk_type String that is either lrisk or trisk.
 #' @return NULL
 validate_input_values <- function(baseline_scenario, shock_scenario, scenario_geography,
```
MirijaHa, line 10: the conditions in this function do not look entirely correct to me,
1. carbon_price_model is a character variable and needs to be added explicitly in the variables on which validate_values_in_values is called (to make sure this is done correctly I would suggest adding a unit test where you use an incorrect carbon tax name)
2. carbon_price_model is not in vector_numeric_args from the start so should not have to be removed from there

MirijaHa (changes requested):

tests/testthat/test-validate.R
frmfisc: styler done

R/read.R
frmfisc: makes sense done

R/read.R
frmfisc: ok done - hopefully at least :)

R/process_data.R
frmfisc: done

R/validate.R
frmfisc: done however problems with the test

frmfisc: so what i dont understand rn is that the validate_values_in_values test fails i have no idea what i am doing wrong - is there a mistake in the function ?

R/validate.R
```diff
@@ -26,7 +26,13 @@ validate_input_values <- function(baseline_scenario, shock_scenario, scenario_ge
     input_args[which(names(input_args) %in% c("carbon_price_model"))] <- NULL
   }
 
-  c("baseline_scenario", "shock_scenario", "scenario_geography") %>%
+  validate_values_in_values <- c("baseline_scenario", "shock_scenario", "scenario_geography","carbon_price_model")
```
MirijaHa, line 29: please use a name like vector_character_args (equivalent to vector_numeric_args below)

MirijaHa (commented):
2 small things.
The reason for the failing unit tests is not so easy to spot. You called the vector that holds the character variables the same name as the function that is used to validate those values. Thus you basically overwrote what validate_values_in_values is, and thus the function no longer gets caught. You can fix it by renaming the vector as I suggested in the comments

MirijaHa (approved):
⚱️

PR #448, opened 2023-02-09
Add ngfs scenarios

frmfisc:
this pr whitelists the new ngfs data and weo data 
additionally validity check of the scenarios is added (might not be the most elegant way) 
snapshot test fails for stress args which makes sense - but for some reason once i try to accept the changes it tells me there is nothing to be updated


frmfisc:
snapshot test läuft - wrapper no problems, szenarien laufen :)

MirijaHa (approved):

PR #450, opened 2023-02-10
Add scenario type

frmfisc:
add a scenario type such that non-allowed baseline shock combinations fail

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -197,3 +197,16 @@ test_that("Error is thrown if colnames are missing", {
     "columns: D, E."
   )
 })
+
+# validate scenario_type-------------------------------------------------------
+test_that("Error is thrown if NGFS scenario is combined with non-NGFS scenario", {
+  data <- tibble::tibble(
+    a = c("NGFS2021_GCAM_CP"),
+    b = c("WEO2021_NZE_2050")
+
+  )
+
+  expect_error(scenario_type <- infer_scenario_type(
+    data = data
+  ))
+})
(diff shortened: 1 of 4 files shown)


tests/testthat/test-utils.R
```diff
@@ -197,3 +197,15 @@ test_that("Error is thrown if colnames are missing", {
...
+# validate scenario_type-------------------------------------------------------
+test_that("Error is thrown if NGFS scenario is combined with non-NGFS scenario", {
+  data <- tibble::tibble(
+    a = c("NGFS2021_GCAM_CP"),
+    b = c("WEO2021_NZE_2050")
+  )
+
+  expect_error(scenario_type <- infer_scenario_type(
+    data = data
+  ))
```
MirijaHa, line 210: pro tip: it is a good idea to add a fraction of the expected error message, because expect_error will also be true if the function fails for other reasons. A la expect_error(bla, "cannot be combined")

tests/testthat/test-utils.R
```diff
@@ -197,3 +197,15 @@ test_that("Error is thrown if colnames are missing", {
     "columns: D, E."
   )
 })
+
+# validate scenario_type-------------------------------------------------------
+test_that("Error is thrown if NGFS scenario is combined with non-NGFS scenario", {
```
MirijaHa, line 202: if you want to be really great also ad a unit test for when it words, i.e. that e.g. "is_ngfs" gets returned

MirijaHa (approved):
just improvement suggestoins, nothing blocking

PR #451, opened 2023-02-15
Ngfs end year lookup

frmfisc:
this pr adds an end_year function that is either 2040 or 2060 depending on the scenario type 
at the moment snap shot tests fail as end_year_arg is carried into the result list. should i prevent this from happening or is this in fact nice to have? cant decide

### DIFF
--- a/R/lookup.R
+++ b/R/lookup.R
@@ -18,8 +18,6 @@ prod_nesting_vars_lookup <- c(
 
 calculation_level_lookup <- "company"
 
-end_year_lookup <- 2040
-
 time_horizon_lookup <- 5
 
 flat_multiplier_lookup <- 1
--- a/R/run_trisk.R
+++ b/R/run_trisk.R
@@ -85,13 +85,17 @@ run_trisk <- function(input_path,
     shock_scenario = shock_scenario
   )
 
+
   args_list$output_path <- customise_output_path(
     output_path = args_list$output_path,
     iter_var = iter_var
   )
 
-  st_results_list <- run_stress_test_iteration(args_list)
+  args_list$end_year <- end_year_lookup(scenario_type = scenario_type)
 
+
+  st_results_list <- run_stress_test_iteration(args_list)
+browser()
   result_names <- names(st_results_list[[1]])
   st_results <- result_names %>%
     purrr::map(function(tib) {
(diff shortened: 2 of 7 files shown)


MirijaHa:
> this pr adds an end_year function that is either 2040 or 2060 depending on the scenario type at the moment snap shot tests fail as end_year_arg is carried into the result list. should i prevent this from happening or is this in fact nice to have? cant decide

If end_year becomes an input argument anyway it is less work to jsut leave it as is now.
Just to go sure, did you check that, apart from the new column, results did not change?

R/utils.R
```diff
@@ -480,3 +480,11 @@ infer_scenario_type <- function(baseline_scenario, shock_scenario) {
     )
   }
 }
+
+end_year_lookup <- function(scenario_type) {
+  return(2040)
```
MirijaHa, line 485: this function will always return 2040 irrespecive of scenario_type as a function ends at the first return statement

MirijaHa (changes requested):

R/utils.R
frmfisc: done i finally understand return

R/lookup.R
```diff
@@ -18,8 +18,6 @@ prod_nesting_vars_lookup <- c(
 
 calculation_level_lookup <- "company"
 
-end_year_lookup <- 2040
```
MirijaHa, line 21: very tiny nit: there is still test_end_year_lookup <- 2040 in test_financial_functions, this can be removed

R/run_trisk.R
```diff
@@ -150,7 +154,7 @@ run_stress_test_iteration <- function(args_list) {
 }
 
 # Avoid R CMD check NOTE: "Undefined global functions or variables"
-globalVariables(c(names(formals(run_trisk)), "iter_var"))
+globalVariables(c(names(formals(run_trisk)), "iter_var", "end_year"))
```
MirijaHa, line 157: why did you do this here and not for lrisk?

MirijaHa (commented):
2 minor things

R/run_trisk.R
frmfisc: adding it - i starting doing it bc of a note that was appearing which vanished once i had done it for trisk

R/lookup.R
frmfisc: done

MirijaHa (approved):

PR #454, opened 2023-02-21
Improve coverage

frmfisc:
this pr aims at improving the coverage by joining financial and production data only over id 
full_trajectory now has 1641160 obs compared to 725840


MirijaHa (approved):
😮

PR #456, opened 2023-02-22
add change to lrisk

frmfisc:


ABuller123 (approved):

R/calculate.R
```diff
@@ -130,11 +130,11 @@ calculate_lrisk_trajectory <- function(input_data_list,
       late_sudden_price = !!rlang::sym(glue::glue("price_{target_scenario}"))
     )
 
-  merge_cols <- c("company_name", "id" = "company_id")
+  merge_cols <- c("id" = "company_id")
 
   full_trajectory <- production_data %>%
     dplyr::inner_join(
-      y = input_data_list$financial_data,
+      y = input_data_list$financial_data %>% dplyr::select(!c(company_name)),
```
maurolepore, line 137: Mmm this looks strange. Where do you define `company_name`? Can you point me to it?

Maybe you need `all_of()`?

maurolepore: I viewed the file and can't see the definition of `company_name` which suggests you're referring to the column `company_name`. I suspect you're trying to avoid a warning what dplyr recently introduced. Let me check what's the best way to handle it. ...

frmfisc: quick context - what is happening here is an inner_join between financial data and production data by company_id, 
 both datasets have the company_name variable, however in order to avoid observations being thrown out bc of discrepancies in the spelling between the company names in the two datasets i want to deselect company_name in the financial dataset. i have taken this approach already for trisk, this just replicates it for lrisk.

maurolepore: Ah, make sense. Thanks.

maurolepore:
``` r
library(dplyr, warn.conflicts = FALSE)

data <- tibble(company_name = "abc", another_column = "def")

data %>% 
  # New approach
  select(!"company_name")
#> # A tibble: 1 × 1
#>   another_column
#>   <chr>         
#> 1 def

data %>% 
  # Same
  select(-"company_name")
#> # A tibble: 1 × 1
#>   another_column
#>   <chr>         
#> 1 def

# The idea comes from the warning here
data %>% 
  # Deprecated
  select(.data$company_name)
#> Warning: Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
#> ℹ Please use `"company_name"` instead of `.data$company_name`
#> # A tibble: 1 × 1
#>   company_name
#>   <chr>       
#> 1 abc
```

<sup>Created on 2023-02-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

maurolepore: Feel free to disregard and continue with what you have -- which seems totally fine. The `c()` is a bit confusing because there is only one column --  but does not harm.

Feel free to merge -- as you understand the context, and now have the privileges.

frmfisc: thanks!!

ABuller123 (approved):

PR #461, opened 2023-04-21
Update workflows

MirijaHa:
Updated workflow for pkgdown and R-CMD-Check.
- For pkgdown I used: use_github_action("pkgdown")
- For r-CMD-Check I used use_github_action("check-full"), manually added a timeout limit and removed some runs on older linux version


maurolepore (approved):
LGTM

maurolepore:
Oh, I see  you used the standard workflow in another PR -- so clearly you considered it. Great!

PR #463, opened 2023-04-24
add failing test

MirijaHa:


maurolepore (commented):
So this is to confirm that the workflow fails when you expect it to fail? If so then it seems to work correctly.

maurolepore (approved):
Agree with Mirja to approve despite the broken checks -- as a way to test some protection settings.

MirijaHa:
close as it was just for testing

PR #467, opened 2023-05-02
Oxford

frmfisc:
oxford data integration
i wanna discuss shortly the snapshot test cause they seem to work for stress test args again !


R/utils.R
```diff
@@ -498,6 +502,10 @@ end_year_lookup <- function(scenario_type) {
     end_year <- as.numeric(2100)
   }
 
+  if (scenario_type == "is_oxford") {
```
MirijaHa, line 505: improvement: you could combine all scenario types that have the same end-year in 1 condition a la if (scenario_type in c(a,b,c)) to make it easier to read

MirijaHa (approved):
LGTM
I retriggered the failiing actions, they look like a temporary server side issue, let us see how it plays out.

PR #474, opened 2023-05-17
Time Frame change for Overshoot Direction

ABuller123:
The timeframe for the overshoot direction is set at 1:5, which takes into account the years from 2021 till 2025. However, we need to consider all 5 years of foreward looking production data, so including 2026. 

I added a simple +1 to the time_frame variable, as time_frame is also used differently in other locations. 

Time Horizon is also changed to 2050, which should be pushed in a different PR into master soon. 

Note: This will affect the snapshot tests, but only to a limited degree. For most companies, it should make no difference.


jakubcervenka (approved):
LGTM

PR #475, opened 2023-06-23
Fix/new geographies

bergalli:


bergalli:
One change that is unclear in the commit messages : a renaming on the scenario geographies of capacity_factors has been removed , to instead be applied in STDataMGMT

bergalli:
Need to be updated with geographies renaming once https://github.com/2DegreesInvesting/STDataMGMT/pull/56 is merged

PR #476, opened 2023-06-30
Rename output files

frmfisc:
adjusts file names for the output files.

### DIFF
--- a/R/run_trisk.R
+++ b/R/run_trisk.R
@@ -92,7 +92,9 @@ run_trisk <- function(input_path,
 
   args_list$output_path <- customise_output_path(
     output_path = args_list$output_path,
-    iter_var = iter_var
+    iter_var = iter_var,
+    shock_scenario = shock_scenario,
+    scenario_geography = scenario_geography
   )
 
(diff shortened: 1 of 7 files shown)


R/run_trisk.R
```diff
@@ -163,7 +167,7 @@ globalVariables(c(names(formals(run_trisk)), "iter_var", "end_year"))
 read_and_process_and_calc <- function(args_list) {
   list2env(args_list, envir = rlang::current_env())
 
-  log_path <- file.path(output_path, paste0("log_file_", iter_var, ".txt"))
+  log_path <- file.path(output_path, paste0("log_file_", iter_var, sep = "_", shock_scenario, sep = "_", scenario_geography, ".txt"))
```
bergalli, line 170: I think it's fine here, but when a name grows too long and uses mixed seperators, string formatting is an option
```r
sprintf("log_file_%s_%s_%s.txt", iter_var, shock_scenario, scenario_geography)
```

frmfisc: thanks

bergalli (approved):

PR #479, opened 2023-08-17
Whitelisting update regions

frmfisc:
snapshot changes as we have new geographies 
whitelisting of new regions


bergalli (dismissed):

bergalli:
Github job can't complete because of a timeout when installing dependancies. Not much to do except rerunning until it passes or removing some dependancies (stringi takes 1 mn to build for example)


Should be investigated why this job takes 15mn while the rest is done in 5

bergalli (dismissed):

bergalli (approved):

PR #481, opened 2023-08-25
Move start year 2022

frmfisc:
moving start year to 2022, we need to keep 2021 around for ScenarioAnalysisinput in the figure of start_year_two; unfortunately some tests fail (check devtools::test())


bergalli (approved):
Fixed the snapshot with 2022

PR #491, opened 2023-09-13
Overshoot clean

frmfisc:
the warning with the uninitalized column unit test coming


frmfisc:
ok should work now - it might be not the most elegant way but its effective hope its clear

bergalli (approved):

PR #506, opened 2023-10-30
change in function

frmfisc:
small change in condition shock year / year

### DIFF
--- a/R/financial_functions.R
+++ b/R/financial_functions.R
@@ -80,7 +80,7 @@ calculate_net_profits_shock_declining_technologies_carbon_tax <- function(data, 
   data_over_shoot_increasing <- data %>% dplyr::filter(.data$overshoot_direction == "Increasing") %>%
     dplyr::mutate(
       production_compensation = .data$late_sudden - .data$baseline,
-      carbon_tax = ifelse(year > shock_year, .data$carbon_tax, 0),
+      carbon_tax = ifelse(year < shock_year, 0, .data$carbon_tax),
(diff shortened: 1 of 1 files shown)


R/financial_functions.R
```diff
@@ -80,7 +80,7 @@ calculate_net_profits_shock_declining_technologies_carbon_tax <- function(data,
   data_over_shoot_increasing <- data %>% dplyr::filter(.data$overshoot_direction == "Increasing") %>%
     dplyr::mutate(
       production_compensation = .data$late_sudden - .data$baseline,
-      carbon_tax = ifelse(year > shock_year, .data$carbon_tax, 0),
+      carbon_tax = ifelse(year < shock_year, 0, .data$carbon_tax),
```
bergalli, line 83: shouldn't it be `<=` ?

frmfisc: its the same in result - i prefer these kind of clear statements

bergalli: it's not the same, with `>` the carbon tax is applied 1 year after the shock year, while with `<` the carbon tax is applied on the shock year included. i.e. 1 year of carbon tax difference.

But then maybe it's better to apply it on the shock year included, meaning your modification corrects a bug ?

frmfisc: i know thats why i change it shockyear is when carbon tax is applied ...

ABuller123 (approved):

R/financial_functions.R
frmfisc: like before 2030 shock year would have meant 2031 has the carbon tax, but we want it to be already applied in 2030

PR #516, opened 2024-01-08
Fix/speed improvements

bergalli:
Major PR that aims to improve the speed of trisk computation by replacing iterative operation by vectorized ones.


bergalli:
continuity test results available here : https://raw.githack.com/2DegreesInvesting/r2dii.climate.stress.test/fix/speed_improvements/data-raw/continuity-test.html

No impact on trajectories
neglictible impact on NPV and PD diff

